PR #5, opened 2019-08-27
use same JWT implementation as we use server-side

tbarbugli:
# Submit a pull request

## CLA

- [ ] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required).
- [ ] The code changes follow best practices

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -1,10 +1,9 @@
 package stream_chat
 
 import (
+	"reflect"
 	"testing"
 	"time"
-
-	"github.com/pascaldekloe/jwt"
 
 	"github.com/stretchr/testify/assert"
 )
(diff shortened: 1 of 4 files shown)


client_test.go
```diff
@@ -48,34 +47,31 @@ func TestNewClient(t *testing.T) {
...
 
-func Test_client_CreateToken(t *testing.T) {
-	c := initClient(t)
-
-	var expire = time.Now().Add(time.Hour)
-	tt := []struct {
-		name   string
+func TestClient_CreateToken(t *testing.T) {
+	type args struct {
+		userId string
```
ferhatelmas: it's consistent throughout the project but it would be a lint error 

`userID` would be better but should be done separately everywhere if intended.

client_test.go
```diff
@@ -48,34 +47,31 @@ func TestNewClient(t *testing.T) {
...
+		wantErr bool
 	}{
-		{"token without expire", time.Time{}},
-		{"token with expire", expire},
+		{"simple without expiration",  args{"tommaso", time.Time{}}, []byte("eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VyX2lkIjoidG9tbWFzbyJ9.v-x-jt3ZnBXXbQ0GoWloIZtVnat2IE74U1a4Yuxd63M"), false},
+		{"simple with expiration",  args{"tommaso", time.Unix(1566941272, 0)}, []byte("eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE1NjY5NDEyNzIsInVzZXJfaWQiOiJ0b21tYXNvIn0.jF4ZbAIEuzS2jRH0uiu3HW9n0NHwT96QkzGlywcG9HU"), false},
 	}
-
-	for _, test := range tt {
-		test := test
```
ferhatelmas: it's nice for scope linting but it's more an issue of tooling than code.

client_test.go
```diff
@@ -48,34 +47,31 @@ func TestNewClient(t *testing.T) {
...
-				expiresIn = jwt.NewNumericTime(test.expire)
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			c, _ := NewClient("key", []byte("secret"))
+			got, err := c.CreateToken(tt.args.userId, tt.args.expire)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("createToken() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if !reflect.DeepEqual(got, tt.want) {
```
ferhatelmas: It's simple but could be even shorter by `testify`. Since it is in vendor, why not use it?

client_test.go
```diff
@@ -48,34 +47,31 @@ func TestNewClient(t *testing.T) {
...
+				t.Errorf("createToken() got = %v, want %v", string(got), tt.want)
 			}
-
-			assert.Equal(t, expiresIn, claims.Expires)
-			assert.Equal(t, testUsers[0].ID, claims.Set["user_id"])
 		})
 	}
-}
+}
```
ferhatelmas: this might be a lack of gofmt

client_test.go
pascaldekloe: The plan is to drop testify. It was left in there to save time during the rushed launch.

ferhatelmas: Sounds good to me then.

For my own learning, what is the reason to drop?

pascaldekloe: What I dislike most is the unreadable errors it produces. To make things even worse, the framework makes the description optional. Spending 30 seconds on a good Errorf description goes a long way.

ferhatelmas: To add from my experience, if diffing long strings and there is a mismatch, it shows no diff. 

It's generally very confusing because at the exact same place, sometimes you see an error and sometimes it's eaten according to the inputs.

PR #8, opened 2019-08-29
Bugfix/race condition

Bogdan-D:
Multiple requests—actually the responses—may update the channel state concurrently.

# Submit a pull request

## CLA

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -324,8 +324,8 @@ func (c *Client) CreateChannel(chanType string, chanID string, userID string, da
 	return ch, err
 }
 
-// todo: cleanup this
-func (ch *Channel) refresh() error {
+// Reload updates channel data from server
+func (ch *Channel) Reload() error {
 	options := map[string]interface{}{
 		"watch":    false,
 		"state":    true,
--- a/channel.go
+++ b/channel.go
@@ -106,7 +106,9 @@ func (ch *Channel) query(options map[string]interface{}, data map[string]interfa
 		data = map[string]interface{}{}
 	}
 
-	data["created_by"] = map[string]interface{}{"id": ch.CreatedBy.ID}
+	if ch.CreatedBy != nil {
+		data["created_by"] = map[string]interface{}{"id": ch.CreatedBy.ID}
+	}
 
 	payload["data"] = data
 
@@ -136,7 +138,14 @@ func (ch *Channel) Update(options map[string]interface{}, message string) error 
 
 	p := path.Join("channels", url.PathEscape(ch.Type), url.PathEscape(ch.ID))
 
-	return ch.client.makeRequest(http.MethodPost, p, nil, payload, nil)
+	var resp channelQueryResponse
+	if err := ch.client.makeRequest(http.MethodPost, p, nil, payload, &resp); err != nil {
+		return err
+	}
+
+	ch.update(resp)
+
+	return nil
 }
 
 // Delete removes the channel. Messages are permanently removed.
@@ -150,7 +159,15 @@ func (ch *Channel) Delete() error {
 func (ch *Channel) Truncate() error {
 	p := path.Join("channels", url.PathEscape(ch.Type), url.PathEscape(ch.ID), "truncate")
 
-	return ch.client.makeRequest(http.MethodPost, p, nil, nil, nil)
+	var resp channelQueryResponse
+
+	if err := ch.client.makeRequest(http.MethodPost, p, nil, nil, &resp); err != nil {
+		return err
+	}
+
+	ch.update(resp)
+
+	return nil
 }
 
 // AddMembers adds members with given user IDs to the channel
@@ -173,6 +190,7 @@ func (ch *Channel) AddMembers(userIDs ...string) error {
 	}
 
 	ch.update(resp)
+
 	return nil
 }
 
@@ -219,6 +237,7 @@ func (ch *Channel) AddModerators(userIDs ...string) error {
 	}
 
 	ch.update(resp)
+
 	return nil
 }
 
@@ -241,6 +260,7 @@ func (ch *Channel) DemoteModerators(userIDs ...string) error {
 	}
 
 	ch.update(resp)
+
 	return nil
 }
 
--- a/channel.go
+++ b/channel.go
@@ -9,9 +9,6 @@ import (
 	"time"
 )
 
-// global lock storage for channels
-var channelsLock sync.Map
-
 type ChannelRead struct {
 	User     *User     `json:"user"`
 	LastRead time.Time `json:"last_read"`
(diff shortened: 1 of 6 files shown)


channel.go
```diff
@@ -50,27 +54,39 @@ type Channel struct {
...
 	Channel  *Channel         `json:"channel,omitempty"`
 	Messages []*Message       `json:"messages,omitempty"`
 	Members  []*ChannelMember `json:"members,omitempty"`
 	Read     []*ChannelRead   `json:"read,omitempty"`
 }
 
-func (q queryResponse) updateChannel(ch *Channel) {
+func (ch *Channel) update(q channelQueryResponse) {
+	// atomically add or create new lock in channel lock store
+	lock, _ := channelsLock.LoadOrStore(ch.CID, &sync.Mutex{})
```
pascaldekloe: Yikes! 😅
```go
ch.mutex.Lock()
refer ch.mutex.Unlock()
```

You could dedupe Channel instantiation (with a sync.Map) if really needed.

stream_chat.go
```diff
@@ -49,6 +49,7 @@ type StreamChannel interface {
 	GetReplies(parentID string, options map[string][]string) (replies []*Message, err error)
 	MarkRead(userID string, options map[string]interface{}) error
 	RemoveMembers(userIDs ...string) error
+	Reload() error
```
pascaldekloe: This is a temporary workaround for the lacking synchronization facilities. Document that.

tbarbugli:
please share more information about the problem. ie: a race output or just point out where we need to sync access

channel.go
Bogdan-D: if lock will be stored in Channel struct, updating channel will be more complex

Bogdan-D:
@tbarbugli there are methods which updates internal channel state from response, if they will be called concurrently - client could get race condition

tbarbugli:
@Bogdan-D yes that was obvious; channel state response is not a type one should share outside the controller, that type is just to return a response. Is there any place where we use it on different go-routines? If not then this is not something we should merge (if yes the solution might not even be to add synchronization)

Bogdan-D:
@tbarbugli we don't have any goroutines here, but clients could use it in any way they want, so they will be surprised about race conditions may appear (an example i've found from imgur, they are using go sdk in http handler and in such case they can't control concurrent requests)

pascaldekloe:
You're supposed to document concurrency features explicitly like “Multiple goroutines may invoke methods on a Conn simultaneously.” in the net package.

channel.go
pascaldekloe: I don't understand how placing the muted in the respective struct causes updates to be more complex? Just remove the map and you're good to go.

channel.go
```diff
@@ -52,28 +49,26 @@ type Channel struct {
 	LastMessageAt time.Time `json:"last_message_at"`
 
 	client *Client
+	lock   *sync.Mutex
```
pascaldekloe: No need for a pointer; just embed the struct.

channel.go
```diff
@@ -315,6 +310,14 @@ func (ch *Channel) UnBanUser(targetID string, options map[string]string) error {
 	return ch.client.UnBanUser(targetID, options)
 }
 
+// newChannel initializes channel with required private fields
```
pascaldekloe: Include the required fields (like ID) and dedupe (with a sync.Map LoadOrStore). This will prevent bugs on the client side.

channel.go
```diff
@@ -52,28 +49,26 @@ type Channel struct {
...
 	Read     []*ChannelRead   `json:"read,omitempty"`
 }
 
-func (ch *Channel) update(q channelQueryResponse) {
-	// atomically add or create new lock in channel lock store
-	lock, _ := channelsLock.LoadOrStore(ch.CID, &sync.Mutex{})
-
-	lock.(*sync.Mutex).Lock()
-	defer lock.(*sync.Mutex).Unlock()
+func (ch *Channel) update(q channelResponse) {
```
pascaldekloe: Not an update; apply(Response).

channel.go
Bogdan-D: it's not public method; what means dedupe?

channel.go
Bogdan-D: there is need for a pointer

Bogdan-D: there is need pointer

channel.go
pascaldekloe: Users can instantiate the same channel twice with the API. A simple deduplication check in combination with your thread safe implementation ensures updates are apply synchronously.

channel.go
Bogdan-D: `ch.update` sounds more logical then `ch.apply`

stream_chat.go
Bogdan-D: I'm not sure if it's temporary and will server-side SDK ever have event listeners :-D

channel.go
pascaldekloe: Update is what you receive; apply is what the fuction does. So applyResponse makes clear the response is applied to ch. Update can mean anything, like change q, mention the presense of an update or whatever. Not that important though. Up to you.

channel.go
pascaldekloe: I don't see how pointers prevent deadlocks?

stream_chat.go
pascaldekloe: Yes, I'm working hard on that (with a full-blown event stream). Should be Q4.

channel.go
Bogdan-D: if you use pointer - after `ch.update` mutex will be pointed to same object. So if other goroutine does waits for `Lock()` at the moment - it will be released after `Unlock()` happened. If we don't use pointer - after `ch.update` mutex will be another object and `Lock` will never released

pascaldekloe: I still don't get it. The mutex applies/protects the Channel it lives in. They map 1 to 1. So when ch.update changes ch, then it uses the mutex embedded in ch, and only that one.

Bogdan-D: https://github.com/GetStream/stream-chat-go/pull/8/files/a2340def2b75ef2f430ebfbade450d75b9ec944f..99b1adc4830130d92a2efc9db9586144bc3ecbac#diff-861e77702f49a71d32be4b3c7600818fR68

Bogdan-D: https://github.com/GetStream/stream-chat-go/pull/8/files/a2340def2b75ef2f430ebfbade450d75b9ec944f..99b1adc4830130d92a2efc9db9586144bc3ecbac#diff-861e77702f49a71d32be4b3c7600818fR68

Bogdan-D: what's your problem with pointer?

pascaldekloe: It allows trickery. An embedded  struct enforces correct use. It documents the 1 to 1 relation.

Bogdan-D:
this will not be merged for now

PR #11, opened 2019-09-10
CreateChannel: channel id is optional when members are present

Bogdan-D:
# Submit a pull request

## CLA

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -268,7 +268,7 @@ func (c *Client) CreateChannel(chanType string, chanID string, userID string, da
 	case chanType == "":
 		return nil, errors.New("channel type is empty")
 	case chanID == "" && !membersPresent:
-		return nil, errors.New("channel ID or members are empty")
+		return nil, errors.New("either channel ID or members must be provided")
 	case userID == "":
 		return nil, errors.New("user ID is empty")
 	}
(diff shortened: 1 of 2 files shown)


channel.go
```diff
@@ -259,11 +259,16 @@ func (ch *Channel) UnBanUser(targetID string, options map[string]string) error {
...
+		_, membersPresent = data["members"]
+	}
+
 	switch {
 	case chanType == "":
 		return nil, errors.New("channel type is empty")
-	case chanID == "":
-		return nil, errors.New("channel ID is empty")
+	case chanID == "" && !membersPresent:
+		return nil, errors.New("channel ID or members are empty")
```
tbarbugli: should be `either channel ID or members must be provided` or `channel ID and members are empty`

channel.go
```diff
@@ -259,11 +259,16 @@ func (ch *Channel) UnBanUser(targetID string, options map[string]string) error {
 
 // CreateChannel creates new channel of given type and id or returns already created one
 func (c *Client) CreateChannel(chanType string, chanID string, userID string, data map[string]interface{}) (*Channel, error) {
+	var membersPresent bool
+	if len(data) > 0 {
+		_, membersPresent = data["members"]
```
ferhatelmas: this line is enough because nil map is accessible by default value

https://play.golang.org/p/HykDxMpGcUe

ferhatelmas: updated at #13

Issue #12, opened 2019-09-10
Proposal: integrate golangci-lint to automate quality checks


ferhatelmas:
If decided, happy to take care it.

tbarbugli:
never used; I kind of like review-dog + metalinter, needs some tweaks to reduce noise but eventually you get nice PR review comments for free

ferhatelmas:
@tbarbugli pushed a PR that does main integration, and corrected issues from errcheck. For the rest, I will wait for your confirmation.

Bogdan-D:
metalinter is dead, golang-ci is better replacement. We'll add it soon

ferhatelmas:
@Bogdan-D see https://github.com/GetStream/stream-chat-go/pull/14 - waiting for review.

PR #14, opened 2019-09-11
Integrate golangci-lint, impi and shfmt

ferhatelmas:
# Submit a pull request

## CLA

**golangi-lint**: most comprehensive metalinter, that
does errcheck, static analysis, etc.

**impi**: ensures import ordering

**shfmt**: automatic shell script formatting similar to gofmt

Add linting to travis.

fixes: #12

### DIFF
--- a/hack/run-lint.sh
+++ b/hack/run-lint.sh
@@ -1,0 +1,61 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+[[ -n ${DEBUG:-} ]] && set -x
+
+gopath="$(go env GOPATH)"
+
+if ! [[ -x "$gopath/bin/golangci-lint" ]]; then
+	echo >&2 'Installing golangci-lint'
+	curl --silent --fail --location \
+		https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b "$gopath/bin" v1.18.0
+fi
+
+# configured by .golangci.yml
+"$gopath/bin/golangci-lint" run
+
+install_impi() {
+	impi_dir="$(mktemp -d)"
+	trap 'rm -rf -- ${impi_dir}' EXIT
+
+	GOPATH="${impi_dir}" \
+		GOBIN="${gopath}/bin" \
+		go get github.com/pavius/impi/cmd/impi
+}
+
+# install impi that ensures import grouping is done consistently
+if ! [[ -x "${gopath}/bin/impi" ]]; then
+	echo >&2 'Installing impi'
+	install_impi
+fi
+
+"$gopath/bin/impi" \
+	--local github.com/GetStream/stream-chat-go \
+	--scheme stdThirdPartyLocal \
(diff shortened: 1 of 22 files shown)


ferhatelmas:
This PR doesn't resolve #12 yet because there are still many issues that need to be fixed so that another PR can have a clean code but I wanted to show if the changes are inline with what we want.

pascaldekloe:
Looks fine apart for the differ removals. Why did you do that? Now tests leave garbage on assertion failure, no?

ferhatelmas:
@pascaldekloe `defer` could fail and they weren't checked but now they are checked (detected by `errcheck`). Otherwise, there shouldn't be any changes regarding messages.

pascaldekloe:
Sure, that error check is fine. But you shouldn't eliminate the defers; create inner functions for that.

ferhatelmas:
> But you shouldn't eliminate the defers; create inner functions for that.

Yes, that could be another option for implementation. Since these are tests, it seemed to me that moving these calls to later seemed more explicit.

If you want the other way, I can update.

ferhatelmas:
@pascaldekloe I will update this part as you said and will fix the rest of the issues, too. Later, today, I will update.

ferhatelmas:
One question is that how about changing package name to `stream` or `streamchat` because underscore in a package name is discouraged.

Since there is no nested package, it should be safe to do comparatively.

pascaldekloe:
@Bogdan-D, you are in charge here ☝️. I think @ferhatelmas is right here.

ferhatelmas:
This should be ready by now. 

If we want package name change, I can update but for now, I added boilerplate comment to silence the issue.

Bogdan-D:
the problem is it will be unexpected for existing clients if package name will change

hack/run-lint.sh
```diff
@@ -0,0 +1,61 @@
...
+set -euo pipefail
+
+[[ -n ${DEBUG:-} ]] && set -x
+
+gopath="$(go env GOPATH)"
+
+if ! [[ -x "$gopath/bin/golangci-lint" ]]; then
+	echo >&2 'Installing golangci-lint'
+	curl --silent --fail --location \
+		https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b "$gopath/bin" v1.18.0
```
Bogdan-D: please use install path from official url

hack/run-lint.sh
```diff
@@ -0,0 +1,61 @@
...
+	install_impi
+fi
+
+"$gopath/bin/impi" \
+	--local github.com/GetStream/stream-chat-go \
+	--scheme stdThirdPartyLocal \
+	--skip "stream_chat_easyjson.go" \
+	./...
+
+install_shfmt() {
```
Bogdan-D: why do we install shfmt?

Bogdan-D:
Merged https://github.com/GetStream/stream-chat-go/pull/19

PR #15, opened 2019-09-16
Feature/bulk mute

Bogdan-D:
# Submit a pull request

## CLA

### DIFF
--- a/user.go
+++ b/user.go
@@ -36,7 +36,26 @@ type User struct {
 // Create a mute
 // targetID: the user getting muted
 // userID: the user muting the target
-func (c *Client) MuteUser(targetIDs []string, userID string) error {
+func (c *Client) MuteUser(targetID string, userID string) error {
+	switch {
+	case len(targetID) == 0:
+		return errors.New("target ID is empty")
+	case userID == "":
+		return errors.New("user ID is empty")
+	}
+
+	data := map[string]interface{}{
+		"target_id": targetID,
+		"user_id":   userID,
+	}
+
+	return c.makeRequest(http.MethodPost, "moderation/mute", nil, data, nil)
+}
+
+// Create a mute
+// targetID: the user getting muted
+// userID: the user muting the target
+func (c *Client) MuteUsers(targetIDs []string, userID string) error {
 	switch {
 	case len(targetIDs) == 0:
 		return errors.New("target IDs are empty")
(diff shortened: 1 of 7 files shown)


user.go
```diff
@@ -71,6 +90,25 @@ func (c *Client) UnmuteUser(targetID string, userID string) error {
 	return c.makeRequest(http.MethodPost, "moderation/unmute", nil, data, nil)
 }
 
+// Removes a mute
+// targetID: the user getting un-muted
+// userID: the user muting the target
+func (c *Client) UnmuteUsers(targetIDs []string, userID string) error {
+	switch {
```
tbarbugli: can't use a switch here (unless you use the fall through but that's ugly)

tbarbugli: (both params needs validation)

user.go
```diff
@@ -71,6 +90,25 @@ func (c *Client) UnmuteUser(targetID string, userID string) error {
...
+	case len(targetIDs) == 0:
+		return errors.New("target IDs is empty")
+	case userID == "":
+		return errors.New("user ID is empty")
+	}
+
+	data := map[string][]string{
+		"target_ids": targetIDs,
+		"user_id":    {userID},
```
tbarbugli: why are you sending `user_id` as []string? either server-side we do something weird or this client is weird

user.go
```diff
@@ -52,13 +52,32 @@ func (c *Client) MuteUser(targetID string, userID string) error {
...
+}
+
 // Removes a mute
 // targetID: the user getting un-muted
 // userID: the user muting the target
 func (c *Client) UnmuteUser(targetID string, userID string) error {
 	switch {
-	case targetID == "":
-		return errors.New("target ID is empty")
+	case len(targetID) == 0:
```
tbarbugli: previous check was better

user.go
Bogdan-D: @tbarbugli `url.Values` does have signature `map[string][]string`

user.go
Bogdan-D: don't remember why this was changed

user.go
Bogdan-D: switch looks cleaner than many if's, but it depends on reviewer

Bogdan-D: but ofc its not problem to just write multiple `if`.

PR #17, opened 2019-09-19
Missing api calls

Bogdan-D:
# Submit a pull request

## CLA

Implementation of missing methods:
1. GetApp, UpdateAppSettings
2. UpdateUser
3. ReactivateUser
4. GetMessage
5. Flag\Unflag message
6. UpdateChannelType
7. VerifyWebhook
8. channel.Query
9. Search

### DIFF
--- a/app.go
+++ b/app.go
@@ -75,7 +75,7 @@ type Policy struct {
 	UpdatedAt time.Time `json:"updated_at"`
 }
 
-type AppResponseFields struct {
+type AppConfig struct {
 	Name                 string                    `json:"name"`
 	OrganizationName     string                    `json:"organization"`
 	PushNotifications    PushNotificationFields    `json:"push_notifications"`
@@ -89,11 +89,11 @@ type AppResponseFields struct {
 }
 
 type appResponse struct {
-	App *AppResponseFields `json:"app"`
+	App *AppConfig `json:"app"`
 }
 
 // GetApp returns app settings
-func (c *Client) GetApp() (*AppResponseFields, error) {
+func (c *Client) GetApp() (*AppConfig, error) {
 	var resp appResponse
 
 	err := c.makeRequest(http.MethodGet, "app", nil, nil, &resp)
--- a/app.go
+++ b/app.go
@@ -92,8 +92,8 @@ type appResponse struct {
 	App *AppConfig `json:"app"`
 }
 
-// GetApp returns app settings
-func (c *Client) GetApp() (*AppConfig, error) {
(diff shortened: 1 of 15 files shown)


app.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	return resp.App, nil
+}
+
+// UpdateAppSettings makes request to update app settings
+// Example of usage:
+//  settings := NewAppSettings().SetDisableAuth(true)
+//  err := client.UpdateAppSettings(settings)
+func (c *Client) UpdateAppSettings(settings *AppSettings) error {
```
pascaldekloe: Update means you replace the settings with the provided object. With all that pointer use, I assume you're sending a patch here?

```go
// PatchAppConfig requests field changes for each non nil value.
func (c *Client) PatchAppConfig(patch  *AppConfigPatch) error {
```

query_test.go
```diff
@@ -40,3 +40,29 @@ func TestClient_QueryChannels(t *testing.T) {
...
+func TestClient_Search(t *testing.T) {
+	c := initClient(t)
+	ch := initChannel(t, c)
+
+	user1, user2 := randomUser(), randomUser()
+
+	text := randomString(10)
+
+	msg1, err := ch.SendMessage(&Message{Text: text + " " + randomString(25)}, user1.ID)
+	mustNoError(t, err)
```
pascaldekloe: Labels! 😡

app.go
```diff
@@ -0,0 +1,113 @@
...
+	DisableAuth          bool                      `json:"disable_auth_checks"`
+	DisablePermissions   bool                      `json:"disable_permissions_checks"`
+}
+
+type appResponse struct {
+	App *AppResponseFields `json:"app"`
+}
+
+// GetApp returns app settings
+func (c *Client) GetApp() (*AppResponseFields, error) {
```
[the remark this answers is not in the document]
Bogdan-D: Don't agree, method names should be about the same on all SDK's

app.go
Bogdan-D: same as above

app.go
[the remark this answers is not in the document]
Bogdan-D: I'd renamed AppResponseFields -> AppConfig; I think good solution woulg be to name it as `GetAppConfig()`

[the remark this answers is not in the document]
pascaldekloe: In terms of documentation that makes sense.

app.go
pascaldekloe: Then document the behaviour.

PR #24, opened 2019-10-24
add support for invite members after channel creation

thesyncim:
# Submit a pull request

## CLA


channel.go
```diff
@@ -189,6 +189,21 @@ func (ch *Channel) AddModerators(userIDs ...string) error {
 	return ch.client.makeRequest(http.MethodPost, p, nil, data, nil)
 }
 
+// InviteMembers invites users with given IDs to the channel
+func (ch *Channel) InviteMembers(userIDs ...string) error {
+	if len(userIDs) == 0 {
+		return errors.New("user IDs are empty")
```
pascaldekloe: `return nil`

thesyncim: im following the same convention used here https://github.com/GetStream/stream-chat-go/pull/24/files#diff-861e77702f49a71d32be4b3c7600818fR210

thesyncim:
> Can you also update interface at stream_chat.go?

yes. we should add a check var _ ChatIntrr

channel.go
pascaldekloe: We define conventions now? 😆

PR #25, opened 2019-10-29
message filters in search request

ravlio:
# Submit a pull request

## CLA


query_test.go
```diff
@@ -49,20 +49,42 @@ func TestClient_Search(t *testing.T) {
...
 
-	got, err := c.Search(SearchRequest{Query: text, Filters: map[string]interface{}{
-		"members": map[string][]string{
-			"$in": {user1.ID, user2.ID},
-		},
-	}})
+	t.Run("search message text", func(t *testing.T) {
+		got, err := c.Search(SearchRequest{Query: text, Filters: map[string]interface{}{
+			"members": map[string][]string{
+				"$in": {user1.ID, user2.ID},
```
pascaldekloe: Search for user IDs in "search message text"?

Bogdan-D: filters are required in any way

query.go
```diff
@@ -105,9 +105,9 @@ func (c *Client) QueryChannels(q *QueryOption, sort ...*SortOption) ([]*Channel,
 
 type SearchRequest struct {
 	// Required
-	Query   string                 `json:"query"`
-	Filters map[string]interface{} `json:"filter_conditions"`
-
+	Query          string                 `json:"query"`
+	Filters        map[string]interface{} `json:"filter_conditions"`
+	MessageFilters map[string]interface{} `json:"message_filters"`
```
Bogdan-D: are message filters required or optional?

ravlio: good point. At least one of the fields is required. Validator.v9 has troubles with cross-field validation, so I've added checking inside the code.

query.go
```diff
@@ -125,6 +126,9 @@ type searchMessageResponse struct {
 func (c *Client) Search(request SearchRequest) ([]*Message, error) {
 	var buf strings.Builder
 
+	if request.Query == "" && request.Filters == nil && request.MessageFilters == nil {
```
pascaldekloe: You probably want a `len(request.Filters) == 0 && len(request.MessageFilters) == 0` here. Len works with the nil value—no need for an extra check.

query_test.go
Bogdan-D: probably members can be replaced with channel id

query.go
ferhatelmas: We can lint this.

keyneston:
@Bogdan-D is this held up just due to failed builds? Which I'm guessing is due to the cleanup issue I've otherwise fixed. If so I can rebase and merge.

PR #26, opened 2019-10-31
system message on channel operations

ravlio:
…tags

# Submit a pull request

## CLA

- [*] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required).
- [*] The code changes follow best practices
- [*] Code changes are tested (add some information if not applicable)

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -111,10 +111,13 @@ func (ch *Channel) query(options map[string]interface{}, data map[string]interfa
 //
 // options: the object to update the custom properties of this channel with
 // message: optional update message
-func (ch *Channel) Update(options map[string]interface{}, message string) error {
+func (ch *Channel) Update(options map[string]interface{}, message *Message) error {
 	payload := map[string]interface{}{
-		"data":    options,
-		"message": message,
+		"data": options,
+	}
+
+	if message != nil {
+		payload["message"] = message
 	}
 
 	p := path.Join("channels", url.PathEscape(ch.Type), url.PathEscape(ch.ID))
@@ -137,7 +140,7 @@ func (ch *Channel) Truncate() error {
 }
 
 // AddMembers adds members with given user IDs to the channel
-func (ch *Channel) AddMembers(userIDs ...string) error {
+func (ch *Channel) AddMembers(userIDs []string, message *Message) error {
 	if len(userIDs) == 0 {
 		return errors.New("user IDs are empty")
 	}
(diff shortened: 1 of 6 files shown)


ravlio:
@Bogdan-D I had to change `time.Time` to `*time.Time` because of these fields are read-only and cause a `message.created_at is a reserved field, message.updated_at is a reserved field` error, and there is no easy way to omit zero time.Time in JSON result if only not to fix it in the JSON codec.

tbarbugli:
I always find it very ugly to pass nil to skip optional parameters (Python got this right 15 years ago...) I usually end up writing two versions of the same function. @pascaldekloe @pascaldekloe any idea?

The fact that we have to use *Message to avoid sending the zero value for some reserved field sounds pretty bad. @Bogdan-D is it really like this on the Go client?

Bogdan-D:
@tbarbugli for me it's also pretty ugly but I haven't found best way to avoid that yet, as I also don't like to create `struct` parameters for each function. You proposal to make it as different function sounds pretty good

ravlio:
@tbarbugli @Bogdan-D tomorrow we'll add another argument which will force us to refactor methods or add other ones, breaking the API. In my opinion, RPC-style methods for client APIs are universal and straightforward.  You do not need more than one argument — just because you can't have more than one body in post message. Please check Consul API: https://godoc.org/github.com/hashicorp/consul/api almost every method has only one main struct-like argument. Or look at the Protobuf — same thing. Check also https://github.com/googleapis/googleapis for the inspiration.

Bogdan-D:
@ravlio @tbarbugli 
I think next approach is the best:
* Required parameters should be inside function signature
* Optional parameters should be used inside different funtion, suffixed by `With`

For example:
* Members are required:
```AddMembers(members []string)```
* Message is optional
```AddMembersWithMessage(members []string, message *Message)```

It looks great. The problem is when there are multiple optional parameters.
We can't add new method for each combination of optional parameters, it's crazy.

So better contept could be like this one:

```
type AddMembersOption interface{
     addMembersOption()
}

// now Message is suitable options for AddMembersWith call
func(m  *Message) addMembersOption() {}

AddMembersWith(members []string, options ...Option)
```

So now there can be any amount of options in one call, we can choose which of our types suitable as option to use with AddMembers. 
Adding new types doesn't break API.

PR #33, opened 2019-11-21
options for several methods

ravlio:
Allow passing a message when:

promoting  / demoting moderators

ban / unban a user (API call only, not for /ban /unban commands)

invite user

accept invite

reject invite

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -77,6 +77,19 @@ func (q queryResponse) updateChannel(ch *Channel) {
 	}
 }
 
+type InviteMembersOptions struct {
+	// Optional system message to
+	Message *Message
+}
+
+type AddModeratorsOptions struct {
+	Message *Message
+}
+
+type DemoteModeratorsOptions struct {
+	Message *Message
+}
+
 // query makes request to channel api and updates channel internal state
 func (ch *Channel) query(options map[string]interface{}, data map[string]interface{}) (err error) {
 	payload := map[string]interface{}{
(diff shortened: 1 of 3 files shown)


ravlio:
@Bogdan-D @pascaldekloe I did not miss a point, please consider my variant more like a proposal. Let's discuss it. Do we really need to copy methods just to provide "optional" options / because Go has no method overloading? There are some straightforward solutions, from rpc-like to Options pattern https://github.com/tmrts/go-patterns/blob/master/idiom/functional-options.md

What do you think?

tbarbugli:
@ravlio I think all other API clients accept a message parameter, not an option parameter; I also think we should not create an option type for only 1 option. I rather break the library the day we actually need more than just an optional message.

ravlio:
@tbarbugli I'm confused. What to choose? The optional message (as a pointer) was initially. Rolling back to it?

Bogdan-D:
my point is that having methods witch requires to pass `nil` all the time is ugly. I haven't idea how to avoid it best when I've started work on `stream-chat-go`, but now I have clear picture.

@tbarbugli you also told the same in past review, now your point of view is changed?

Bogdan-D:
I like this way the best from easy to use:
https://play.golang.org/p/455sVvyCjVU

the only problem `godoc` has bad support for this yet, to show which options can be passed to function

ravlio:
now should be better @tbarbugli @Bogdan-D

PR #34, opened 2019-12-12
Chat 910 search results sorting

ravlio:
Zendesk ticket: https://getstream.zendesk.com/agent/tickets/1604

We need to add order column for search request, so results could be ordered by customer.

derived from: https://github.com/GetStream/stream-chat-go/pull/25


query_test.go
```diff
@@ -49,19 +49,46 @@ func TestClient_Search(t *testing.T) {
...
+			},
+		}})
 
-	mustNoError(t, err)
+		mustNoError(t, err)
 
-	assert.Len(t, got, 2)
+		assert.Len(t, got, 2)
+		mustNoError(t, err)
+		assert.Len(t, got, 2)
```
ferhatelmas: why to duplicate?

query_test.go
```diff
@@ -49,19 +49,46 @@ func TestClient_Search(t *testing.T) {
...
+				}},
+			MessageFilters: map[string]interface{}{
+				"text":  map[string]interface{}{"$q": text},
+				"color": "green",
+			},
+			Sort: []*SortOption{{Field: "created_at", Direction: -1}},
+		})
+
+		mustNoError(t, err)
+		assert.Len(t, got, 1)
```
ferhatelmas: how is sorting tested with one reply?

query.go
```diff
@@ -126,6 +128,9 @@ type searchMessageResponse struct {
 func (c *Client) Search(request SearchRequest) ([]*Message, error) {
 	var buf strings.Builder
 
+	if request.Query == "" && request.Filters == nil && request.MessageFilters == nil {
+		return nil, errors.New("either query or filters must be provided")
```
ferhatelmas: if `MessageFilters` isn't nil, it will skip this part. Seems behavior is different than error message. Is it intended?

ravlio: This is the idea. Filters (Channel filters) and message filters are filters). Error text is correct.

ferhatelmas: OK, but it might not be obvious to the user from the error message because filters means both. How about

> one of the query, filters (channel) or message filters must be provided

pascaldekloe: So we allow no filters, but deny nil? Note that you can call `len(request.Filters) != 0` on a nil Filters.

query_test.go
```diff
@@ -49,19 +49,46 @@ func TestClient_Search(t *testing.T) {
...
-	}})
+	t.Run("search message text", func(t *testing.T) {
+		got, err := c.Search(SearchRequest{Query: text, Filters: map[string]interface{}{
+			"members": map[string][]string{
+				"$in": {user1.ID, user2.ID},
+			},
+		}})
 
-	mustNoError(t, err)
+		mustNoError(t, err)
```
pascaldekloe: Labels *please*! Even better; stop using that damned framework.

```go
if err != nil {
        t.Fatal("search error:", err)
}
```

query.go
ferhatelmas: Yes, this is very error-prone and length check is a very good idea.

However a *nit*; I think checking `len(someSlice) > 0` is superior to `len(someSlice) != 0`

because a length can be zero or bigger but not equal to zero somehow signals negative is possible.

query_test.go
ferhatelmas: I mark this resolved since independent sort test is added.

query_test.go
```diff
@@ -49,19 +49,63 @@ func TestClient_Search(t *testing.T) {
...
+			MessageFilters: map[string]interface{}{
+				"text": map[string]interface{}{"$q": text},
+			},
+			Sort: []*SortOption{{Field: "created_at", Direction: -1}},
+		})
+
+		mustNoError(t, err)
+		assert.Len(t, got, 2)
+
+		assert.Equal(t, got[0].Text, text+" 2")
```
ferhatelmas: The order of the parameters is wrong. First expected then got.

It mightn't be seen on success but on fail, it will generate a wrong failure message.

ferhatelmas:
@ravlio #25 will be closed then?

PR #38, opened 2019-12-23
Combine same type params

ferhatelmas:
Combine same type parameters.

### DIFF
--- a/utils_test.go
+++ b/utils_test.go
@@ -37,10 +37,10 @@ func randomUser() *User {
 	return testUsers[rand.Intn(len(testUsers)-1)]
 }
 
-func randomString(len int) string {
-	bytes := make([]byte, len)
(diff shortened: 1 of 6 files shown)


utils_test.go
```diff
@@ -37,10 +37,10 @@ func randomUser() *User {
 	return testUsers[rand.Intn(len(testUsers)-1)]
 }
 
-func randomString(len int) string {
-	bytes := make([]byte, len)
-	for i := 0; i < len; i++ {
-		bytes[i] = byte(65 + rand.Intn(25)) //A=65 and Z = 65+25
+func randomString(n int) string {
```
pascaldekloe: Use testing/quick for that.

ferhatelmas: Very good idea, recording in issues to do.

PR #55, opened 2020-03-03
Fix quotation in documentation for GetReactions

keyneston:
The example provided in the documentation for GetReactions is incorrect. I've updated the docs.
Specifically the Options type is of `map[string][]string` which means the parameters have to be strings.

### DIFF
--- a/reaction.go
+++ b/reaction.go
@@ -82,7 +82,7 @@ type reactionsResponse struct {
 }
 
 // GetReactions returns list of the reactions for message with given ID.
-// options: Pagination params, ie {"limit":{10}, "idlte": {10}}
+// options: Pagination params, ie {"limit":{"10"}, "idlte": {"10"}}
 func (ch *Channel) GetReactions(messageID string, options map[string][]string) ([]*Reaction, error) {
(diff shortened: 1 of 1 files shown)


reaction.go
```diff
@@ -82,7 +82,7 @@ type reactionsResponse struct {
 }
 
 // GetReactions returns list of the reactions for message with given ID.
-// options: Pagination params, ie {"limit":{10}, "idlte": {10}}
+// options: Pagination params, ie {"limit":{"10"}, "idlte": {"10"}}
 func (ch *Channel) GetReactions(messageID string, options map[string][]string) ([]*Reaction, error) {
```
tbarbugli: why is options `map[string][]string` and not `map[string]interface{}`?

tbarbugli: @keyneston I understand this is not your code in first place but since we are looking at it, good to find out since it looks at least very weird to do this

keyneston: I’ve got a bit of a growing list of small oddities like this. I’ll try and dig into a few of them, and share it when I’m done with the examples.

PR #57, opened 2020-03-05
Add examples for documentation

keyneston:
These are mostly examples I wrote for the website proper, but I decided they would fit well in here.

There are a few small changes that fall out side of that.

1. I added a small delay in one of the tests before test cleanup. This was due to it intermittently failing causing the tests to fail.
2. I added a `(*Client).Channel` function which makes it easier to get a channel with known values. It doesn't make an API call in and of itself, just gives you an object that can be used for API calls.

### DIFF
--- a/channel_test.go
+++ b/channel_test.go
@@ -89,16 +89,12 @@ func TestChannel_AddMembers(t *testing.T) {
 
 // See https://getstream.io/chat/docs/channel_members/ for more details.
 func ExampleChannel_AddModerators() {
-	var err error
 	channel := &Channel{}
 	newModerators := []string{"bob", "sue"}
 
-	err = channel.AddModerators("thierry", "josh")
-	err = channel.AddModerators(newModerators...)
-	err = channel.DemoteModerators(newModerators...)
-	if err != nil {
-		log.Fatalf("%v", err)
-	}
+	_ = channel.AddModerators("thierry", "josh")
+	_ = channel.AddModerators(newModerators...)
+	_ = channel.DemoteModerators(newModerators...)
 }
 
 func TestChannel_InviteMembers(t *testing.T) {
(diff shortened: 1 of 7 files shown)


keyneston:
The test were failing a good portion of the time when it failed to cleanup a channel or channel type during tests. These errors just caused us to run the tests more which then caused more failed cleanup. Either way we end up with junk floating around in the test account. So I changed the cleanup portions of the test to not fail if an error was encountered.

tbarbugli:
> The test were failing a good portion of the time when it failed to cleanup a channel or channel type during tests. These errors just caused us to run the tests more which then caused more failed cleanup. Either way we end up with junk floating around in the test account. So I changed the cleanup portions of the test to not fail if an error was encountered.

IMO the best solution to this problem is to *not* cleanup at all; have tests create data with random IDs and use it only for the current test. Adding a random prefix/suffix to names and IDs is usually the simplest way to keep stuff readable and unique 👉no cleanup.

keyneston:
@tbarbugli we already are doing the random prefix for the names and IDs. We can speed it up by not doing any deletes at all I guess.

channel_test.go
```diff
@@ -411,3 +422,29 @@ func TestChannel_RejectInvite(t *testing.T) {
...
 }
+
+func ExampleChannel_Update() {
+	// https://getstream.io/chat/docs/channel_permissions/?language=python
+	client := &Client{}
+
+	data := map[string]interface{}{
+		"image":      "https://path/to/image",
+		"created_by": "elon",
+		"roles":      map[string]string{"elon": "admin", "gwynne": "moderator"},
```
thesyncim: roles can be a bit confusing (since you are adding a channel and role is a user property )

keyneston: I'm not sure I understand what you are saying. The example was transliterated from the python. So this is just what the python one does 😁

PR #59, opened 2020-03-18
Version 3 of the API

keyneston:
This is a fairly large modification to the how the API is used. I've tried to increase documentation as I go.

A lot of this is stylistic in an attempt to increase ergonomics of the client. Let me know what you like/don't like.

The biggest change is the addition of the Option type. It was designed so that functions could take a variadic list of Options. It would then compile them as needed. Originally I wanted to make it so that there were sub-groups of options. Say UserDeleteOptions and so on, but that level of management would get too high.

This was inspired by but is really unrelated to internal ticket CHAT-570.


options.go
```diff
@@ -0,0 +1,129 @@
...
+func (o option) Key() string {
+	return o.key
+}
+
+func (o option) Value() interface{} {
+	return o.value
+}
+
+// PaginateGreaterThan returns an Option for paginating.
+func PaginateGreaterThan(id int) Option {
```
tbarbugli: `id string`

options.go
```diff
@@ -0,0 +1,129 @@
...
+	return NewOption("id_gt", id)
+}
+
+// PaginateOffset returns an Option for paginating.
+func PaginateOffset(id int) Option {
+	return NewOption("offset", id)
+}
+
+// PaginateLimit returns an Option for paginating.
+func PaginateLimit(id int) Option {
```
tbarbugli: rename arg to limit

options.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+// PaginateGreaterThan returns an Option for paginating.
+func PaginateGreaterThan(id int) Option {
+	return NewOption("id_gt", id)
+}
+
+// PaginateOffset returns an Option for paginating.
+func PaginateOffset(id int) Option {
+	return NewOption("offset", id)
```
tbarbugli: rename id into offset

query.go
```diff
@@ -8,17 +8,45 @@ import (
...
+// SortField designates a string as being a sortable field. The default sort fields
+// are all declared as constants.
+type SortField = string
+
+const (
+	// SortFieldUserID lets you sort by the user's ID.
+	SortFieldUserID SortField = "user_id"
+
+	// SortFieldLastActive lets you sort by when the user was last active.
+	SortFieldLastActive SortField = "last_active"
```
tbarbugli: add `last_updated` (ask Marcelo)

query.go
```diff
@@ -8,17 +8,45 @@ import (
...
+	SortDescending SortDirection = -1
+)
+
+// SortField designates a string as being a sortable field. The default sort fields
+// are all declared as constants.
+type SortField = string
+
+const (
+	// SortFieldUserID lets you sort by the user's ID.
+	SortFieldUserID SortField = "user_id"
```
tbarbugli: not really useful

query.go
tbarbugli: add last_message_at, created_at, updated_at

ferhatelmas:
Are we fine with package rename? 

I think I prefer new name but anyway wanted to point for confirmation.

keyneston:
@ferhatelmas Tomasso asked me to switch it back. It is better semantics and stuff, but the transition cost might be too high.

channel_type.go
```diff
@@ -21,12 +21,43 @@ const (
 	MessageRetentionForever = "infinite"
 )
 
+// ChannelTypeLabel marks a string as a type of channel. The builtin settings are all
+// available as consts.
+type ChannelTypeLabel = string
```
ferhatelmas: It sounds to me `Label` suffix is a bit unnecessary.

keyneston: Agreed. I had a TODO here about finding a better name, sadly the linter is a overly zealous.

`ChannelTypeLabel` is definitely too much, but we can't use `ChannelType` because well that is already a type/struct. So it needs something, label is the best I could come up with. I'm open to suggestions.

ferhatelmas: btw, why don't we use renamed types ?

keyneston: @ferhatelmas sorry I don't know what you mean

client.go
```diff
@@ -117,14 +118,19 @@ func (c *Client) newRequest(method, path string, params url.Values, data interfa
...
+	params url.Values,
+	data interface{},
+	result easyjson.Unmarshaler,
+) error {
 	r, err := c.newRequest(method, path, params, data)
 	if err != nil {
 		return err
 	}
 
+	log.Printf("%v", r.URL.String())
```
ferhatelmas: is this intentional to keep?

channel_type.go
keyneston: Oh do you mean like https://play.golang.org/p/RIVqIDAOWGN this seems obvious in retrospect

ferhatelmas: Yes because this way, we don't type safety (not sure if we need) and can't add methods if we need later. Either way, it's an improvement since readability is much better.

For naming, I saw later. Within constraints, sounds good tome.

client.go
```diff
@@ -153,12 +171,12 @@ func (c *Client) createToken(params map[string]interface{}, expire time.Time) ([
 
 	claims.Expires = jwt.NewNumericTime(expire.Round(time.Second))
 
-	return claims.HMACSign(jwt.HS256, c.apiSecret)
+	return claims.HMACSign(jwt.HS256, []byte(c.apiSecret))
```
ferhatelmas: might set it as byte slice in initialization

go.mod
```diff
@@ -1,9 +1,9 @@
-module github.com/GetStream/stream-chat-go/v2
+module github.com/GetStream/stream-chat-go/v3
 
 go 1.12
 
 require (
 	github.com/getstream/easyjson v0.0.0-20190923162548-5adc7b1f33fa
 	github.com/pascaldekloe/jwt v1.5.0
-	github.com/stretchr/testify v1.3.0
+	github.com/stretchr/testify v1.4.0
```
ferhatelmas: while updating, maybe for https://pkg.go.dev/github.com/stretchr/testify@v1.5.1?tab=doc

helpers.go
```diff
@@ -0,0 +1,11 @@
...
+
+// String is a helper function for making pointers to strings.
+func String(i string) *string {
+	return &i
+}
+
+// Bool is a helper function for making pointers to bools.
+func Bool(i bool) *bool {
+	return &i
+}
```
ferhatelmas: These are very simple, why do we need to provide them?

Also, they aren't used.

options.go
```diff
@@ -0,0 +1,129 @@
...
+	optionKeyTargetUserID = "target_user_id"
+	optionKeyTimeout      = "timeout"
+	optionKeyLocation     = "url"
+	optionKeyReason       = "reason"
+)
+
+func compileOptions(opts ...Option) url.Values {
+	val := url.Values{}
+	for _, opt := range opts {
+		if opt == nil {
```
ferhatelmas: might reverse the logic to not nil

options.go
```diff
@@ -0,0 +1,129 @@
...
+
+	return val
+}
+
+func optionAsString(o Option) string {
+	switch v := o.Value().(type) {
+	case string:
+		return v
+	default:
+		return fmt.Sprintf("%v", v)
```
ferhatelmas: `fmt.Sprint(v)`

ferhatelmas: actually we don't need switch

options.go
```diff
@@ -0,0 +1,129 @@
...
+	return option{key: key, value: value}
+}
+
+// makeTargetID is a helper function for making a target ID option.
+func makeTargetID(targetID string) Option {
+	return NewOption(optionKeyTargetUserID, targetID)
+}
+
+// OptionTimeout returns a timeout option for banning or muting a user. If the
+// time is less than one second then
```
ferhatelmas: not complete

query.go
```diff
@@ -8,17 +8,45 @@ import (
...
+
+	// SortFieldLastActive lets you sort by when the user was last active.
+	SortFieldLastActive SortField = "last_active"
+)
+
+// SortOption is used to configure a sort field.
 type SortOption struct {
-	Field     string `json:"field"`     // field name to sort by,from json tags(in camel case), for example created_at
-	Direction int    `json:"direction"` // [-1, 1]
+	// Field is the naem of the field to sort by. This should be snake case.
```
ferhatelmas: typo `naem`

run-lint.sh
```diff
@@ -9,7 +9,7 @@ gopath="$(go env GOPATH)"
 if ! [[ -x "$gopath/bin/golangci-lint" ]]; then
 	echo >&2 'Installing golangci-lint'
 	curl --silent --fail --location \
-		https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b "$gopath/bin" v1.22.2
+		https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b "$gopath/bin" v1.23.8
```
ferhatelmas: `1.24.0` ?

stream_chat.go
```diff
@@ -1,7 +1,7 @@
 //go:generate go run github.com/getstream/easyjson/easyjson -pkg -all
 
-// Package stream_chat provides chat via stream API
-package stream_chat // nolint: golint
+// package stream_chat //nolint:golint provides "chat as an API" via stream
```
ferhatelmas: something is wrong here

user.go
```diff
@@ -110,107 +110,141 @@ func (c *Client) UnmuteUsers(targetIDs []string, userID string) error {
...
+func (b *banUserInput) AddOptions(options []Option) {
+	for _, opt := range options {
+		switch opt.Key() {
+		case optionKeyTimeout:
+			seconds := opt.Value().(int)
+			b.Timeout = &seconds
+		case optionKeyReason:
+			reason := optionAsString(opt)
+			b.Reason = &reason
+		}
```
ferhatelmas: not sure if we should fail on unknown keys to let users that they are doing something wrong

helpers.go
keyneston: Whoops this was for a previous version that used AWS api semantics. They aren't needed here because we aren't using them, but they are really useful because you can't take the pointer of a static value.

```
// e.g.
foo.bar = &false // <- This doesn't work
foo.bar = aws.Bool(false) // <- this does
```

stream_chat.go
keyneston: Ohh good catch, not sure why go test didn't fail catch this.

PR #60, opened 2020-03-19
implement channel mute

thesyncim:
# Submit a pull request

## CLA

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,8 +43,8 @@ import (
 	stream "github.com/GetStream/stream-chat-go/v2"
 )
 
-var APIKey = os.Getenv("STREAM_API_KEY")
-var APISecret = os.Getenv("STREAM_API_SECRET")
+var APIKey = os.Getenv("STREAM_CHAT_API_KEY")
+var APISecret = os.Getenv("STREAM_CHAT_API_SECRET")
 var userID = "" // your server user id
 
 func main() {
--- a/channel.go
+++ b/channel.go
@@ -491,7 +491,7 @@ func (ch *Channel) Mute(userID string, expiration *time.Duration) error {
 	}
 
 	if expiration != nil {
-		*expiration = *expiration / time.Millisecond
+		*expiration /= time.Millisecond
 	}
 
 	data := map[string]interface{}{
--- a/channel.go
+++ b/channel.go
@@ -485,6 +485,37 @@ func (ch *Channel) RejectInvite(userID string, message *Message) error {
 	return ch.client.makeRequest(http.MethodPost, p, nil, data, nil)
 }
 
+func (ch *Channel) Mute(userID string, expiration *time.Duration) error {
+	if userID == "" {
+		return errors.New("user ID must be not empty")
+	}
+
+	if expiration != nil {
+		*expiration = *expiration / time.Millisecond
+	}
+
+	data := map[string]interface{}{
+		"user_id":     userID,
+		"channel_cid": ch.CID,
+		"expiration":  expiration,
+	}
+
+	return ch.client.makeRequest(http.MethodPost, "moderation/mute/channel", nil, data, nil)
+}
+
+func (ch *Channel) Unmute(userID string) error {
+	if userID == "" {
+		return errors.New("user ID must be not empty")
+	}
+
+	data := map[string]interface{}{
+		"user_id":     userID,
+		"channel_cid": ch.CID,
+	}
+
+	return ch.client.makeRequest(http.MethodPost, "moderation/unmute/channel", nil, data, nil)
+}
+
 //nolint: godox
 // todo: cleanup this
 func (ch *Channel) refresh() error {
--- a/channel.go
+++ b/channel.go
@@ -485,9 +485,9 @@ func (ch *Channel) RejectInvite(userID string, message *Message) error {
 	return ch.client.makeRequest(http.MethodPost, p, nil, data, nil)
 }
 
-func (ch *Channel) Mute(userID string, expiration *time.Duration) error {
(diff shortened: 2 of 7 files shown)


query.go
```diff
@@ -77,11 +80,12 @@ type queryChannelResponseData struct {
 func (c *Client) QueryChannels(q *QueryOption, sort ...*SortOption) ([]*Channel, error) {
 	qp := queryChannelRequest{
 		State:            true,
-		FilterConditions: q,
+		FilterConditions: q.Filter,
```
ferhatelmas: what happens to pagination?

query.go
```diff
@@ -77,11 +80,12 @@ type queryChannelResponseData struct {
...
 		Sort:             sort,
+		UserID:           q.UserID,
 	}
 
-	data, err := easyjson.Marshal(&qp)
+	data, err := json.Marshal(&qp)
```
ferhatelmas: why is there a change here ?

user.go
```diff
@@ -31,7 +45,8 @@ type User struct {
 
 	ExtraData map[string]interface{} `json:"-,extra"` //nolint: staticcheck
 
-	Mutes []*Mute `json:"mutes,omitempty"`
+	Mutes       []*Mute        `json:"mutes,omitempty"`
+	ChannelMute []*ChannelMute `json:"channel_mutes,omitempty"`
```
ferhatelmas: I would suggest following existing convention of `Mutes` so `ChannelMutes`.

channel_test.go
```diff
@@ -423,6 +423,56 @@ func TestChannel_RejectInvite(t *testing.T) {
...
+		members = append(members, testUsers[i].ID)
+	}
+
+	ch, err := c.CreateChannel("messaging", randomString(12), serverUser.ID, map[string]interface{}{
+		"members": members,
+	})
+	require.NoError(t, err, "create channel")
+
+	// mute the channel
+	mute, err := ch.Mute(members[0], nil)
```
ferhatelmas: maybe expiration test too

README.md
```diff
@@ -43,8 +43,8 @@ import (
 	stream "github.com/GetStream/stream-chat-go/v2"
 )
 
-var APIKey = os.Getenv("STREAM_API_KEY")
-var APISecret = os.Getenv("STREAM_API_SECRET")
+var APIKey = os.Getenv("STREAM_CHAT_API_KEY")
+var APISecret = os.Getenv("STREAM_CHAT_API_SECRET")
```
ferhatelmas: :+1:

query.go
thesyncim: we don't need to use easyjson at all here.
the previous code was broken

query.go
thesyncim: the previous code didn't work.
we should probably fix that. maybe in another PR
filter_conditions must be map[string]interface{}

query.go
ferhatelmas: yes, not needed but sometimes std sometime easyjson can be error prone. 
Since we can't drop it because we need it in some cases, it might be better to stick to it.

What was broken actually?

query.go
thesyncim: nothing :).
this was broken.
we should fix/ add that in another PR

ferhatelmas: OK, but how do you pass pagination now?

thesyncim: we dont support paggination for queryChannels

query.go
thesyncim: see https://github.com/GetStream/stream-chat-go/pull/60#discussion_r395514655

query.go
ferhatelmas: OK, existing code isn't working and error prone since enables passing wrong parameters but actually not supported. 

It would be nicer to get the bug fixed first, though.

thesyncim: agree

query.go
ferhatelmas: Broken part is actually not marshaling but the type. When above change (`q -> q.filter`), is this still needed?

> we can change to EasyJSON (no strong opinions on this)

I think sticking to one would be better.

channel.go
```diff
@@ -485,6 +485,43 @@ func (ch *Channel) RejectInvite(userID string, message *Message) error {
 	return ch.client.makeRequest(http.MethodPost, p, nil, data, nil)
 }
 
+func (ch *Channel) Mute(userID string, expiration *time.Duration) (*ChannelMuteResponse, error) {
+	if userID == "" {
+		return nil, errors.New("user ID must be not empty")
+	}
+
+	if expiration != nil {
+		*expiration /= time.Millisecond
```
keyneston: This is technically correct, but we shouldn't be doing it. We should be creating a new non-`time.Duration` variable and use that instead of converting a `time.Duration` in place and making it no longer a real `time.Duration`. Also `int(expiration.Seconds())` should do what is needed here.

user.go
```diff
@@ -9,13 +9,27 @@ import (
 	"time"
 )
 
+// Mute represents an user mute
```
keyneston:
```suggestion
// Mute represents a user mute
```

[Source](https://www.writersdigest.com/online-editor/a-before-consonants-and-an-before-vowels-is-not-the-rule). `an`/`a` is decided by the sound the letter makes. (Which gets confusing because in the US it would be `an herb` vs UK `a herb`.

query.go
keyneston: EasyJSON makes sense for internal stuff where we need to optimism performance. For external stuff, api clients, and the like I'd put my vote for avoiding non-standard dependencies like this.

user.go
thesyncim: thanks

ferhatelmas:
@thesyncim this is already integrated, closing

PR #62, opened 2020-03-24
Switch to using mailru/easyjson

### LINKED ISSUE #61: Package name inconsistencies create issues on Windows
`go: 1.13.6`
`os: windows`

Windows treats `GetStream` and `getstream` as the same thing. When vendoring the library on windows and deploying to linux-based systems (i.e. via a Google Cloud function), the vendoring process will use `github.com/GetStream` for all packages.

However, there is a dependency on `github.com/getstream/easyjson` (which uses the lowercase convention) and the build tools will fail to find the vendored dependency when interacting with case-sensitive file systems / OSes.

It would be great if the package names were consistent to avoid this issue.

keyneston:
This requires we no longer use our special `extra` directive. Instead we need to use the new `easyjson.UnknownFields` interface.

Fixes #61


keyneston:
I tried having an `ExtraData` struct that we embedded but that really didn't seem to want to work. Not consistently at least. Copy pasting the functions into all the places we needed it works better.

go.mod
```diff
@@ -3,7 +3,8 @@ module github.com/GetStream/stream-chat-go/v2
 go 1.12
 
 require (
```
ferhatelmas: also time to bump this

message.go
```diff
@@ -1,11 +1,15 @@
...
 
 import (
+	"encoding/json"
 	"errors"
 	"net/http"
 	"net/url"
 	"path"
 	"time"
+
+	jlexer "github.com/mailru/easyjson/jlexer"
```
ferhatelmas: why rename ?

keyneston: Not sure, it is what go imports did by default, and I didn't fight it.

message.go
```diff
@@ -43,7 +47,29 @@ type Message struct {
 	UpdatedAt *time.Time `json:"updated_at,omitempty"`
 
 	// any other fields the user wants to attach a message
-	ExtraData map[string]interface{}
+	ExtraData map[string]interface{} `json:"-"`
```
ferhatelmas: How about having a custom type `ExtraData` that implements these interfaces and other types embeds it. Can we prevent duplication ?

go.mod
keyneston: Do you mean `go 1.12`? If so I'd rather we err on the side of allowing the user to decide.

message.go
keyneston: No, I tried this but it gets all wonky, it is also a backwards incompatible change.

go.mod
ferhatelmas: if we are testing with go1.12, that's fine but if we stop, we should bump so that user can get a nice message.

message.go
ferhatelmas: > backwards incompatible change

Yes.

With a version bump should be fine though I believe. Maybe with v3 series.

message.go
ferhatelmas: It's unneeded, might be outdated goimports.

PR #73, opened 2020-04-20
Use go1.14

ferhatelmas:
* test with go1.14 (aligned with go versioning)
* bump min required to go1.13 (error wrapping can be used)

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -6,7 +6,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        goVer: [1.12, 1.13]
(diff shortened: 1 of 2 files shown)


ferhatelmas:
@keyneston it might be a breaking change for some users. Do you think it's fine with v2.3.0?

keyneston:
@ferhatelmas hmm, you are right. I need to finish doing the cleanup for version 3.0 if you would prefer to roll this into that.

ferhatelmas:
It might be better that way and this can wait for sure so no rush, at your convenience, when you're done we can merge this.

ferhatelmas:
closing because added new versions but kept old ones, no breaking change.

PR #76, opened 2020-04-29
Channel Mutes & Cleanup Query Options

### LINKED ISSUE #74: [BUG] QueryChannel pagination limit incorrect implementation
Hi people

According to the source code, pagination parameters for the channel querying should be passed within the `filter_conditions`. 
In docs it is suggested to pass the query parameters on the root level of the payload JSON structure (`queryChannelRequest` struct in Go SDK instead of `QueryOption` nested struct).

**References**
In Go SDK:
https://github.com/GetStream/stream-chat-go/blob/master/query.go#L102

In docs:
https://getstream.io/chat/docs/query_channels/#query-options

**Steps to reproduce**
1. Create at least two channels
2. Call `QueryChannels` method without any pagination options passed
3. Make sure you receive your channels in the response
4. Call `QueryChannels` method with explicit pagination rules (limit: 10, offset: 1)

_Expected_
Channels are returned respecting the provided pagination parameters

_Actual_
Empty channels list is returned

keyneston:
This is a fork of #60 as @thesyncim had already started down this path. I just took it a step further and did some additional cleanup of query options.

This should also fix #74

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -491,7 +491,7 @@ func (ch *Channel) Mute(userID string, expiration *time.Duration) error {
 	}
 
 	if expiration != nil {
-		*expiration = *expiration / time.Millisecond
+		*expiration /= time.Millisecond
 	}
 
 	data := map[string]interface{}{
--- a/channel.go
+++ b/channel.go
@@ -490,14 +490,12 @@ func (ch *Channel) Mute(userID string, expiration *time.Duration) (*ChannelMuteR
 		return nil, errors.New("user ID must be not empty")
 	}
 
-	if expiration != nil {
-		*expiration /= time.Millisecond
-	}
-
 	data := map[string]interface{}{
 		"user_id":     userID,
 		"channel_cid": ch.CID,
-		"expiration":  expiration,
+	}
+	if expiration != nil {
+		data["expiration"] = int(expiration.Seconds())
 	}
 
 	var mute = &ChannelMuteResponse{}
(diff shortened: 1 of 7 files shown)


keyneston:
Rebased onto master.

channel.go
```diff
@@ -485,6 +485,41 @@ func (ch *Channel) RejectInvite(userID string, message *Message) error {
...
+
+	data := map[string]interface{}{
+		"user_id":     userID,
+		"channel_cid": ch.CID,
+	}
+	if expiration != nil {
+		data["expiration"] = int(expiration.Seconds())
+	}
+
+	var mute = &ChannelMuteResponse{}
```
ferhatelmas, line 501: tiny but prefer `:=` here

reasoning: https://dave.cheney.net/practical-go/presentations/qcon-china.html#_use_a_consistent_declaration_style

user.go
```diff
@@ -13,13 +13,27 @@ import (
...
+type ChannelMute struct {
+	User      User       `json:"user"`
+	Channel   Channel    `json:"channel"`
+	Expires   *time.Time `json:"expires"`
+	CreatedAt time.Time  `json:"created_at"`
+	UpdatedAt time.Time  `json:"updated_at"`
+}
+
+type ChannelMuteResponse struct {
+	ChannelMutes ChannelMute `json:"channel_mute"`
```
ferhatelmas, line 34: this is a single object so `ChannelMute` would be more correct

keyneston:
@ferhatelmas I think I fixed all the things you pointed out. Can you reapprove?

PR #77, opened 2020-05-01
Attempt to add pagination functions

keyneston:
This is an experiment at adding paging functions for querying. It is similar to what I've seen from AWS and similar libraries. It just abstracts the looping and what not from you.

Example:
```go
	results := []*chat.Channel{}

	err = client.QueryChannelsPages(&chat.QueryOption{
		Filter: map[string]interface{}{
			"members": map[string]interface{}{"$in": []string{"555"}},
		},
	}, func(more []*chat.Channel) bool {
		results = append(results, more...)
		return true
	})

	if err != nil {
		panic(err)
	}
```

Thoughts on the idea, and style? It already is helping me with some of the scripts I'm writing for support purposes.

### DIFF
--- a/query.go
+++ b/query.go
@@ -2,6 +2,7 @@ package stream_chat // nolint: golint
 
 import (
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"net/url"
 	"strings"
@@ -131,6 +132,38 @@ func (c *Client) QueryChannels(q *QueryOption, sort ...*SortOption) ([]*Channel,
 	}
 
 	return result, err
+}
+
+type ChannelPaginationFunc func([]*Channel) bool
+
+func (c *Client) QueryChannelsPages(q *QueryOption, paginationFunc ChannelPaginationFunc) error {
+	if q == nil {
+		return fmt.Errorf("what should be done here?")
+	}
+
+	// Make a copy of q so we can manipulate it without affecting memory outside of here
+	newQ := QueryOption{}
+	newQ = *q
+
+	limit := 30
+
+	for i := 0; ; i++ {
+		newQ.Limit = limit
+		newQ.Offset = limit * i
+
+		res, err := c.QueryChannels(&newQ)
+		if err != nil {
+			return err
+		}
+		if len(res) == 0 {
+			return nil
+		}
+
+		// If the paginationFunc indicates we should not continue then we stop.
+		if ok := paginationFunc(res); !ok {
+			return nil
+		}
+	}
 }
 
 type SearchRequest struct {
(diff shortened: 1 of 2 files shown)


query.go
```diff
@@ -133,6 +134,38 @@ func (c *Client) QueryChannels(q *QueryOption, sort ...*SortOption) ([]*Channel,
 	return result, err
 }
 
+type ChannelPaginationFunc func([]*Channel) bool
+
+func (c *Client) QueryChannelsPages(q *QueryOption, paginationFunc ChannelPaginationFunc) error {
+	if q == nil {
+		return fmt.Errorf("what should be done here?")
```
ferhatelmas, line 141: defaults, IMHO

query.go
```diff
@@ -133,6 +134,38 @@ func (c *Client) QueryChannels(q *QueryOption, sort ...*SortOption) ([]*Channel,
...
+	}
+
+	// Make a copy of q so we can manipulate it without affecting memory outside of here
+	newQ := QueryOption{}
+	newQ = *q
```
ferhatelmas, line 146: why not `newQ := *q` directly?

query.go
```diff
@@ -133,6 +134,38 @@ func (c *Client) QueryChannels(q *QueryOption, sort ...*SortOption) ([]*Channel,
...
+
+	limit := 30
```
ferhatelmas, line 148: it should care given limit maybe

query.go
```diff
@@ -133,6 +134,38 @@ func (c *Client) QueryChannels(q *QueryOption, sort ...*SortOption) ([]*Channel,
...
+
+	for i := 0; ; i++ {
+		newQ.Limit = limit
+		newQ.Offset = limit * i
```
ferhatelmas, line 152: also caring for the starting offset, maybe

ferhatelmas:
we might alias it to `type M = map[string]interface{}`

ferhatelmas:
For naming, `client.PageQueryChannels` sounds better to me than `client.QueryChannelsPages`.

keyneston:
@ferhatelmas thanks for the feedback!

> For naming, client.PageQueryChannels sounds better to me than client.QueryChannelsPages.

Definitely agree.

> we might alias it to type M = map[string]interface{} to make it shorter.

In scripty code I often do `type obj map[string]interface{}`, I tend to keep libraries, prod code, and etc to the full name. That said I do really like `logrus.Fields` which is just an alias. Maybe:

```go
type Filter map[string]interface{}
```
Or something?

ferhatelmas:
Very welcome. A bit more context for the alias suggestion.

My inspiration for alias comes from web frameworks:

[echo has Map](https://godoc.org/github.com/labstack/echo#Map)(was M before), [gin has H](https://godoc.org/github.com/gin-gonic/gin#H) for the same type.

Since it's common for a JSON response, they wanted to make it shorter for the repeated pattern I guess.

But in their case, it's not an alias since they need to attach some methods for different kinds of serialization.

PR #84, opened 2020-06-25
Drop easyjson

ferhatelmas:
in favor of standard library. It does double encoding both ways.


tbarbugli:
did not check the diff yet but I love the PR description :)

PR #86, opened 2020-07-23
User mute expiration

gumuz:
Added options to `MuteUser` and `MuteUsers` to support the recently added `timeout` option for Mute expiration

### DIFF
--- a/query.go
+++ b/query.go
@@ -2,7 +2,6 @@ package stream_chat // nolint: golint
 
 import (
 	"encoding/json"
-	"fmt"
 	"net/http"
 	"net/url"
 	"strings"
@@ -80,8 +79,6 @@ func (c *Client) QueryUsers(q *QueryOption, sorters ...*SortOption) ([]*User, er
 	if err != nil {
 		return nil, err
 	}
-
-	fmt.Println(string(data))
 
 	values := make(url.Values)
 	values.Set("payload", string(data))
--- a/user.go
+++ b/user.go
@@ -15,10 +15,11 @@ import (
 
 // Mute represents a user mute.
 type Mute struct {
-	User      User      `json:"user"`
-	Target    User      `json:"target"`
-	CreatedAt time.Time `json:"created_at"`
-	UpdatedAt time.Time `json:"updated_at"`
+	User      User       `json:"user"`
+	Target    User       `json:"target"`
+	CreatedAt time.Time  `json:"created_at"`
+	UpdatedAt time.Time  `json:"updated_at"`
+	Expires   *time.Time `json:"expires"`
 }
 
 // ChannelMute represents a channel mute.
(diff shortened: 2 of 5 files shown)


gumuz:
Not sure if it fails because of a bug or some configuration problem. @ferhatelmas is this test dependent on the fixes you were preparing for the chat api?

user.go
```diff
@@ -77,39 +78,37 @@ func (u User) MarshalUnknowns(out *jwriter.Writer, first bool) {
...
 		return errors.New("target IDs are empty")
 	case userID == "":
 		return errors.New("user ID is empty")
 	}
 
-	data := map[string]interface{}{
-		"target_ids": targetIDs,
-		"user_id":    userID,
-	}
+	options["target_ids"] = targetIDs
```
ferhatelmas, line 108: `options` can be nil here

user_test.go
```diff
@@ -30,31 +30,41 @@ func TestClient_MuteUser(t *testing.T) {
...
 	assert.NotEmpty(t, mute.Target, "mute has target")
+	assert.NotEmpty(t, mute.Expires, "mute has expires")
 }
 
 func TestClient_MuteUsers(t *testing.T) {
 	c := initClient(t)
 	initChannel(t, c)
 
-	users := []string{randomUser().ID, randomUser().ID}
+	usernames := []string{randomUser().ID, randomUser().ID}
```
ferhatelmas, line 54: this randomness can cause issues in tests once in a while, plug in some specific users

user_test.go
```diff
@@ -30,31 +30,41 @@ func TestClient_MuteUser(t *testing.T) {
...
+	err := c.MuteUsers(usernames, serverUser.ID, map[string]interface{}{"timeout": 60})
 	require.NoError(t, err, "mute user")
+
+	users, err := c.QueryUsers(&QueryOption{
+		Filter: map[string]interface{}{
+			"id": map[string]string{"$eq": serverUser.ID},
+		}})
+	require.NoError(t, err, "query users")
+
+	for _, mute := range users[0].Mutes {
```
ferhatelmas, line 65: ensure length before, it can panic here otherwise

query.go
```diff
@@ -81,8 +80,6 @@ func (c *Client) QueryUsers(q *QueryOption, sorters ...*SortOption) ([]*User, er
 		return nil, err
 	}
 
-	fmt.Println(string(data))
```
ferhatelmas, line 84: :+1:

ferhatelmas:
Put some comments, see them. 

However, it is failing because tests put some garbage and didn't clean it so it's at size limit of some resources. I will clean them and additionally we should update tests to clean up themselves.

ferhatelmas:
I did some cleanup in its app, should be fine now.

user_test.go
```diff
@@ -28,33 +28,61 @@ func TestClient_MuteUser(t *testing.T) {
 	c := initClient(t)
 	initChannel(t, c)
 
-	user := randomUser()
+	assert.GreaterOrEqualf(t, len(testUsers), 1, "not enough test users: %+v", testUsers)
```
ferhatelmas, line 31: it won't stop the test if fails so you can still have panics so if it's a hard requirement like length check, prefer `require`.

user_test.go
```diff
@@ -28,33 +28,63 @@ func TestClient_MuteUser(t *testing.T) {
...
+	assert.NotEmpty(t, mute.Target, "mute should have a Target")
+	assert.NotEmpty(t, mute.Expires, "mute should have Expires")
 }
 
 func TestClient_MuteUsers(t *testing.T) {
 	c := initClient(t)
 	initChannel(t, c)
 
-	users := []string{randomUser().ID, randomUser().ID}
+	assert.GreaterOrEqualf(t, len(testUsers), 2, "not enough test users: %+v", testUsers)
```
ferhatelmas, line 72: same logic should be here too

gumuz: I was just about to push that now!

ferhatelmas: ah, sorry, I thought it was ready

ferhatelmas:
@gumuz Do you want to release it ?

* changelog
* bump version 
* tag (minor in this case)

PR #87, opened 2020-07-29
Updated version and changelog for 2.4.0

gumuz:
@ferhatelmas I've also tagged it but I'm not sure about the process to release it, how would I do that?


ferhatelmas:
Pushing the tag at this stage is good then create a release with the existing tag. You can add the relevant part from changelog to release description too.

ferhatelmas:
This is correct because if you push the tag before, version will be old one.

PR #95, opened 2020-09-23
[CHAT-1159] import channel messages

thesyncim:
# Submit a pull request

## CLA

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -104,7 +104,7 @@ func (q queryResponse) updateChannel(ch *Channel) {
 	}
 }
 
-type SendBulkMessagesResponse struct {
+type ImportChannelMessagesResponse struct {
 	Messages []Message `json:"messages"`
 }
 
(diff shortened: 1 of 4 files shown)


channel.go
```diff
@@ -104,6 +104,10 @@ func (q queryResponse) updateChannel(ch *Channel) {
 	}
 }
 
+type SendBulkMessagesResponse struct {
+	Messages []Message `json:messages`
```
gumuz, line 108: I think it should be json:"messages", like the linter suggests

thesyncim: ouch :D

gumuz (approved):

channel.go
```diff
@@ -212,6 +216,27 @@ func (ch *Channel) RemoveMembers(userIDs []string, message *Message) error {
 	return nil
 }
 
+// ImportChannelMessages is a batch endpoint for inserting multiple messages.
+func (ch *Channel) ImportChannelMessages(messages ...*Message) (*ImportChannelMessagesResponse, error) {
```
ferhatelmas, line 220: could you add it to channel interface ?

ferhatelmas (commented):
see one comment otherwise good

channel.go
thesyncim: done

ferhatelmas: thanks but we need to rename it to `ImportMessages`, it's already a method of channel

thesyncim: good point, fixing it

ferhatelmas (approved):
LGTM

PR #107, opened 2020-12-02
MessageLimit and MemberLimit should be pointers

ruudniew:
So we don't automatically omit 0, but also don't automatically set the limits to 0 if they're not specified


ferhatelmas (approved):
LGTM

keyneston:
@ruudniew is this ready to be merged?

ferhatelmas:
yup

PR #115, opened 2021-03-08
add GetRateLimits endpoint support

ruggi:
# Submit a pull request

## CLA

This PR introduces support for the `GetRateLimits` endpoint, which returns the current rate limits quotas and usage.

The params are thought so that if they are empty/zero they won't be included in the call and the whole list of limits will be returned.

### DIFF
--- a/rate_limits.go
+++ b/rate_limits.go
@@ -34,40 +34,77 @@ type GetRateLimitsResponse struct {
 	Web        RateLimitsMap `json:"web,omitempty"`
 }
 
-// GetRateLimitsOptions configures the Client.GetRateLimits call.
-type GetRateLimitsOptions struct {
-	// ServerSide, if true, restricts the returned limits to server-side clients only.
-	ServerSide bool
-	// Android, if true, restricts the returned limits to Android clients only.
-	Android bool
-	// IOS, if true, restricts the returned limits to iOS clients only.
-	IOS bool
-	// Web, if true, restricts the returned limits to web clients only.
-	Web bool
-	// Endpoints restricts the returned limits info to the specified endpoints.
-	Endpoints []string
+type getRateLimitsParams struct {
+	serverSide bool
+	android    bool
+	iOS        bool
+	web        bool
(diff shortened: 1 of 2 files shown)


rate_limits.go
```diff
@@ -0,0 +1,80 @@
...
+	IOS bool
+	// Web, if true, restricts the returned limits to web clients only.
+	Web bool
+	// Endpoints restricts the returned limits info to the specified endpoints.
+	Endpoints []string
+}
+
+// GetRateLimits returns the current rate limit quotas and usage. If no params are toggled, all the limits
+// for all platforms are returned.
+func (c *Client) GetRateLimits(options GetRateLimitsOptions) (GetRateLimitsResponse, error) {
```
ferhatelmas, line 53: That is fine but in general, having a struct here is making harder to change it later.

Having `...options` is better from API perspective.

`GetRateLimits(Android, IOS, Endpoints("query_channels", "query_users")) ..`

ferhatelmas (approved):
LGTM

rate_limits.go
ruggi: I agree on paper with the variadic options, but I think in this case it would be better to keep it as a single struct. The alternative is to have e.g.

```go
type RateLimitParams struct {
...
}

func RateLimitOptions func(p *RateLimitParams)

func GetRateLimits(opts ...RateLimitOptions) {
	// ...
	params := RateLimitParams{}
	for _, opt := range options {
		opt(&params)
	}
	// ...
```

which imho doesn't add much value

ruggi: happy to have the change above ^ if and when there are non-default values in the params

ferhatelmas: Having params non-exported and variadic options enables us to change anything without breaking as we want even if we can't see how from today.
 
```go
type rateLimitParams struct {
...
}
```

However, it's good for now.

ruggi: @ferhatelmas updated

PR #117, opened 2021-03-09
Update GetRateLimits for server change in endpoints

ferhatelmas:


ruggi:
it's ok if the tests are not passing, let's wait for they to be green before merging @ferhatelmas

PR #119, opened 2021-04-19
Add SendUserCustomEvent method

yaziine:
# Submit a pull request

## CLA

This PR adds
- the `SendUserCustomEvent` method for sending custom event to a user
- a new `UserCustomEvent` type

⚠️ This shouldn't be merge until the API integration

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -1,6 +1,7 @@
 package stream_chat // nolint: golint
 
 import (
+	"encoding/json"
 	"testing"
 	"time"
 
@@ -130,4 +131,25 @@ func TestSendUserCustomEvent(t *testing.T) {
 			require.EqualError(t, err, test.expectedErr)
 		})
 	}
+}
+
+func TestMarshalUnmarshalUserCustomEvent(t *testing.T) {
+	ev1 := UserCustomEvent{
+		Type: "custom_event",
+		ExtraData: map[string]interface{}{
+			"name":   "John Doe",
+			"age":    99.0,
+			"hungry": true,
+			"fruits": []interface{}{},
+		},
+	}
+
+	b, err := json.Marshal(ev1)
+	require.NoError(t, err)
+
+	ev2 := UserCustomEvent{}
+	err = json.Unmarshal(b, &ev2)
+	require.NoError(t, err)
+
+	require.Equal(t, ev1, ev2)
 }
--- a/client_test.go
+++ b/client_test.go
@@ -88,3 +88,46 @@ func TestClient_CreateToken(t *testing.T) {
 		})
 	}
 }
+
+func TestSendUserCustomEvent(t *testing.T) {
+	c := initClient(t)
+
+	tests := []struct {
+		name        string
+		event       *Event
+		userID      string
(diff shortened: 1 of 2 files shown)


client_test.go
```diff
@@ -88,3 +88,46 @@ func TestClient_CreateToken(t *testing.T) {
...
+	tests := []struct {
+		name         string
+		event        *UserCustomEvent
+		targetUserID string
+		expectedErr  string
+	}{
+		{
+			name: "ok",
+			event: &UserCustomEvent{
+				Type: "custom_event",
```
ferhatelmas, line 104: could you add some custom data to test marshaling?

event.go
```diff
@@ -121,9 +117,67 @@ func (ch *Channel) SendEvent(event *Event, userID string) error {
...
+
+	ExtraData map[string]interface{} `json:"-"`
+
+	CreatedAt time.Time `json:"created_at,omitempty"`
+}
+
+type userCustomEventForJSON UserCustomEvent
+
+func (e *UserCustomEvent) UnmarshalJSON(data []byte) error {
+	// TODO: merge this method with Event.UnmarshalJSON
```
ferhatelmas, line 144: would be nice if we could address this one too

ferhatelmas: however, I would go that way since they are quite different. That's why just dropping this todo is good addressing to me.

client_test.go
yaziine: I can but I am not sure we have a way to ensure that the data is marshalled correctly. We can only check that `Marshal` doesn't return an error. It's probably enough though.

ferhatelmas: Yup, checking error is better and there could be another test, just exercises marshal/unmarshal exclusively.

ferhatelmas: Another idea is to marshal and unmarshal and check if it was done correctly

ferhatelmas: It seems what you did

client_test.go
```diff
@@ -88,3 +89,66 @@ func TestClient_CreateToken(t *testing.T) {
...
+	}
+
+	b, err := json.Marshal(ev1)
+	require.NoError(t, err)
+
+	ev2 := UserCustomEvent{}
+	err = json.Unmarshal(b, &ev2)
+	require.NoError(t, err)
+
+	require.EqualValues(t, *ev1, ev2)
```
ferhatelmas, line 153: `Equal` is better since we want to see type equality too (even if obvious)

client_test.go
```diff
@@ -88,3 +89,66 @@ func TestClient_CreateToken(t *testing.T) {
...
+	}
+}
+
+func TestMarshalUnmarshalUserCustomEvent(t *testing.T) {
+	ev1 := &UserCustomEvent{
+		Type: "custom_event",
+		ExtraData: map[string]interface{}{
+			"name":   "John Doe",
+			"age":    99.0,
+			"hungry": true,
```
ferhatelmas, line 142: could you use zero value `false` ?

ferhatelmas (approved):
LGTM

PR #125, opened 2021-05-21
Added QueryMessageFlags client method

gumuz:


message.go
```diff
@@ -279,25 +279,78 @@ func (c *Client) DeleteMessage(msgID string) error {
...
 	}
 
+	if userID == "" {
+		return errors.New("user ID is empty")
+	}
+
 	options := map[string]interface{}{
 		"target_message_id": msgID,
+		"user_id":           userID,
 	}
```
gumuz, line 354: these were missing the mandatory userID serverside parameter

channel_type_test.go
```diff
@@ -13,6 +14,8 @@ func prepareChannelType(t *testing.T, c *Client) *ChannelType {
 	ct, err := c.CreateChannelType(ct)
 	require.NoError(t, err, "create channel type")
 
+	time.Sleep(6 * time.Second)
```
gumuz, line 17: I believe there was a latency/timing issue here, where the new ChannelType wasn't directly available to the test that created them

.gitignore
```diff
@@ -14,3 +14,6 @@
 .vscode/
 .idea/
 .envrc
+
+# vendored dependencies and binaries
+vendor/
```
ferhatelmas, line 19: this is actually need since we don't vendor but having doesn't hurt

channel_type_test.go
ferhatelmas: possible

message.go
ferhatelmas: this is good 👍🏻

ferhatelmas (approved):
LGTM

PR #130, opened 2021-06-30
Proxy command list to createTypeRequest when creating channel type

AnatolyRugalev:
Currently commands that set to the `ChannelType` are not proxied to the `channelTypeRequest` when creating a channel type. This PR fixes this


ferhatelmas (approved):
LGTM

PR #135, opened 2021-08-18
Add data_template and server_key to firebase configuration (close #134)

### LINKED ISSUE #134: DataTemplate field missing in FirebaseConfig?
image
Docs show that there should be a `data_template` field, but there isn't one
image

driver-devel:


app.go
```diff
@@ -62,7 +62,9 @@ type APNConfig struct {
 
 type FirebaseConfig struct {
 	Enabled              bool   `json:"enabled"`
+	ServerKey            string `json:"server_key"`
 	NotificationTemplate string `json:"notification_template"`
+	DataTemplate         string `json:"data_template"`
```
ferhatelmas, line 67: good to add `omitempty`

ferhatelmas: normally we should separate request and response structs since they have different fields.

app.go
```diff
@@ -62,7 +62,9 @@ type APNConfig struct {
 
 type FirebaseConfig struct {
 	Enabled              bool   `json:"enabled"`
+	ServerKey            string `json:"server_key"`
 	NotificationTemplate string `json:"notification_template"`
+	DataTemplate         string `json:"data_template,omitempty"`
```
ferhatelmas, line 67: notification too please

ferhatelmas (approved):
LGTM

PR #142, opened 2021-10-19
[CHAT-2457] Include result in the response

yaziine:


yaziine:
Just checked, indeed. Closing this. Will update API. Good catch 👍

PR #148, opened 2021-11-15
[CHAT-2592] ShadowBan & Pin methods

gumuz:
Add ShadowBan, RemoveShadowBan, PinMessage & UnPinMessage methods.

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -441,7 +441,7 @@ func (ch *Channel) UnBanUser(targetID string, options map[string]string) error {
 // userID: user who shadow bans target.
 // options: additional shadow ban options, ie {"timeout": 3600, "reason": "offensive language is not allowed here"}.
 func (ch *Channel) ShadowBan(targetID, userID string, options map[string]interface{}) error {
-	if options == nil{
+	if options == nil {
 		options = map[string]interface{}{}
 	}
 
@@ -453,7 +453,7 @@ func (ch *Channel) ShadowBan(targetID, userID string, options map[string]interfa
 
 // RemoveShadowBan removes the ban for target user ID on this channel.
 func (ch *Channel) RemoveShadowBan(targetID string, options map[string]string) error {
-	if options == nil{
+	if options == nil {
 		options = map[string]string{}
 	}
 
--- a/channel.go
+++ b/channel.go
@@ -437,10 +437,10 @@ func (ch *Channel) UnBanUser(targetID string, options map[string]string) error {
 	return ch.client.UnBanUser(targetID, options)
 }
 
-// ShadowBan bans target userID from this channel
-// userID: user who shadow bans target.
+// ShadowBan shadow bans userID from this channel
+// bannedByID: user who shadow bans userID.
 // options: additional shadow ban options, ie {"timeout": 3600, "reason": "offensive language is not allowed here"}.
-func (ch *Channel) ShadowBan(targetID, bannedByID string, options map[string]interface{}) error {
+func (ch *Channel) ShadowBan(userID, bannedByID string, options map[string]interface{}) error {
 	if options == nil {
 		options = map[string]interface{}{}
 	}
@@ -448,11 +448,11 @@ func (ch *Channel) ShadowBan(targetID, bannedByID string, options map[string]int
 	options["type"] = ch.Type
 	options["id"] = ch.ID
 
-	return ch.client.ShadowBan(targetID, bannedByID, options)
+	return ch.client.ShadowBan(userID, bannedByID, options)
 }
 
 // RemoveShadowBan removes the ban for target user ID on this channel.
-func (ch *Channel) RemoveShadowBan(targetID string, options map[string]string) error {
+func (ch *Channel) RemoveShadowBan(userID string, options map[string]string) error {
 	if options == nil {
 		options = map[string]string{}
 	}
@@ -460,7 +460,7 @@ func (ch *Channel) RemoveShadowBan(targetID string, options map[string]string) e
 	options["type"] = ch.Type
 	options["id"] = ch.ID
 
-	return ch.client.RemoveShadowBan(targetID, options)
+	return ch.client.RemoveShadowBan(userID, options)
 }
 
 // Query fills channel info with state (messages, members, reads).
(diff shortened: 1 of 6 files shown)


channel.go
```diff
@@ -437,6 +437,32 @@ func (ch *Channel) UnBanUser(targetID string, options map[string]string) error {
 	return ch.client.UnBanUser(targetID, options)
 }
 
+// ShadowBan bans target userID from this channel
+// userID: user who shadow bans target.
+// options: additional shadow ban options, ie {"timeout": 3600, "reason": "offensive language is not allowed here"}.
+func (ch *Channel) ShadowBan(targetID, userID string, options map[string]interface{}) error {
```
mahboubii, line 443: difference between target and userID is not clear

channel.go
```diff
@@ -437,6 +437,32 @@ func (ch *Channel) UnBanUser(targetID string, options map[string]string) error {
...
+	}
+
+	options["type"] = ch.Type
+	options["id"] = ch.ID
+
+	return ch.client.ShadowBan(targetID, userID, options)
+}
+
+// RemoveShadowBan removes the ban for target user ID on this channel.
+func (ch *Channel) RemoveShadowBan(targetID string, options map[string]string) error {
```
mahboubii, line 455: instead of targetID maybe better to just use userID?

message.go
```diff
@@ -269,6 +273,60 @@ func (c *Client) UpdateMessage(msg *Message, msgID string) (*Message, error) {
 	return resp.Message, nil
 }
 
+// PartialUpdateMessage partially updates message with given msgID.
+func (c *Client) PartialUpdateMessage(msgID, userID string, updates PartialUpdate) (*Message, error) {
```
mahboubii, line 277: we can make it clear here that userID is optional field

message.go
```diff
@@ -269,6 +273,60 @@ func (c *Client) UpdateMessage(msg *Message, msgID string) (*Message, error) {
...
+	var resp messageResponse
+
+	p := path.Join("messages", url.PathEscape(msgID))
+
+	options := map[string]interface{}{
+		"Set":   updates.Set,
+		"Unset": updates.Unset,
+		"User": map[string]string{
+			"id": userID,
+		},
```
mahboubii, line 294: would API throw for empty/nil `user.id`?

message.go
mahboubii: PartialUpdate also accepts `options{skip_enrich_url: bool}`

channel.go
```diff
@@ -437,6 +437,32 @@ func (ch *Channel) UnBanUser(targetID string, options map[string]string) error {
 	return ch.client.UnBanUser(targetID, options)
 }
 
+// ShadowBan shadow bans userID from this channel
+// bannedByID: user who shadow bans userID.
+// options: additional shadow ban options, ie {"timeout": 3600, "reason": "offensive language is not allowed here"}.
+func (ch *Channel) ShadowBan(userID, bannedByID string, options map[string]interface{}) error {
```
mahboubii, line 443: should we type the options since it only accepts certain fields?

channel.go
```diff
@@ -437,6 +437,32 @@ func (ch *Channel) UnBanUser(targetID string, options map[string]string) error {
...
+	}
+
+	options["type"] = ch.Type
+	options["id"] = ch.ID
+
+	return ch.client.ShadowBan(userID, bannedByID, options)
+}
+
+// RemoveShadowBan removes the ban for target user ID on this channel.
+func (ch *Channel) RemoveShadowBan(userID string, options map[string]string) error {
```
mahboubii, line 455: same here, not clear what options we are accepting

channel.go
gumuz: not sure, makes it less flexible. we use interface maps everywhere

gumuz: @yaziine what do you think?

channel.go
```diff
@@ -437,6 +437,32 @@ func (ch *Channel) UnBanUser(targetID string, options map[string]string) error {
...
+		options = map[string]interface{}{}
+	}
+
+	options["type"] = ch.Type
+	options["id"] = ch.ID
+
+	return ch.client.ShadowBan(userID, bannedByID, options)
+}
+
+// RemoveShadowBan removes the ban for target user ID on this channel.
```
yaziine, line 454:
```suggestion
// RemoveShadowBan removes the shadow ban for target user ID on this channel.
```

channel.go
yaziine: following the discussion above, i think we should make it crystal clear which options are acceptable in the go doc.

But here for `RemoveShadowBan`, i am not sure it is possible to pass any extra options.

message.go
```diff
@@ -269,6 +273,70 @@ func (c *Client) UpdateMessage(msg *Message, msgID string) (*Message, error) {
 	return resp.Message, nil
 }
 
+// PartialUpdateMessage partially updates message with given msgID.
+func (c *Client) PartialUpdateMessage(msgID string, updates PartialUpdate, options map[string]interface{}) (*Message, error) {
+	switch {
+	case len(updates.Set) == 0 && len(updates.Unset) == 0:
+		return nil, errors.New("updates is empty")
```
yaziine, line 280: `errors.New("updates are empty")`  or `errors.New("updates parameter is empty")`

message.go
```diff
@@ -269,6 +273,70 @@ func (c *Client) UpdateMessage(msg *Message, msgID string) (*Message, error) {
 	return resp.Message, nil
 }
 
+// PartialUpdateMessage partially updates message with given msgID.
+func (c *Client) PartialUpdateMessage(msgID string, updates PartialUpdate, options map[string]interface{}) (*Message, error) {
```
yaziine, line 277: Same here which options can be passed to this method?

message.go
```diff
@@ -269,6 +273,70 @@ func (c *Client) UpdateMessage(msg *Message, msgID string) (*Message, error) {
...
+
+// PinMessage pins the message with given msgID.
+func (c *Client) PinMessage(msgID, pinnedByID string, expiration *time.Duration) (*Message, error) {
+	updates := PartialUpdate{
+		Set: map[string]interface{}{
+			"pinned": true,
+		},
+	}
+	if expiration != nil {
+		updates.Set["pin_expires"] = int(expiration.Milliseconds())
```
yaziine, line 315: Why do we need to cast here?

yaziine (approved):

PR #151, opened 2021-12-09
feat: add truncate options

peterdeme:
Add truncate options. Using the splat operator to keep backward compatibility.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,26 @@
+# :recycle: Contributing
+
+Contributions to this project are very much welcome, please make sure that your code changes are tested and that they follow
+Go best-practices.
+
+## Getting started
+
+### Required environmental variables
+
+The tests require at least two environment variables: `STREAM_CHAT_API_KEY` and `STREAM_CHAT_API_SECRET`. There are multiple ways to provide that:
+- simply set it in your current shell (`export STREAM_CHAT_API_KEY=xyz`)
+- you could use [direnv](https://direnv.net/)
+- if you debug the tests in VS Code, you can set up an env file there as well: `"go.testEnvFile": "${workspaceFolder}/.env"`.
+
+### Code formatting & linter
+
+We enforce code formatting with [`gufumpt`](https://github.com/mvdan/gofumpt) (a stricter `gofmt`). If you use VS Code, it's recommended to set this setting there for auto-formatting:
+
+```json
+"editor.formatOnSave": true,
+"gopls": {
+    "formatting.gofumpt": true
+}
+```
+
+Gofumpt will mostly take care of your linting issues as well.
--- a/README.md
+++ b/README.md
@@ -66,5 +66,5 @@ func main() {
 
 ### Contributing
 
-Contributions to this project are very much welcome, please make sure that your code changes are tested and that follow
-Go best-practices.
+Contributions to this project are very much welcome, please make sure that your code changes are tested and that they follow
+Go best-practices. You can find some tips in [CONTRIBUTING.md](./CONTRIBUTING.md).
--- a/channel.go
+++ b/channel.go
@@ -183,11 +183,52 @@ func (ch *Channel) Delete() error {
 	return ch.client.makeRequest(http.MethodDelete, p, nil, nil, nil)
 }
 
+type truncateOptions struct {
+	HardDelete  bool       `json:"hard_delete,omitempty"`
+	SkipPush    bool       `json:"skip_push,omitempty"`
+	TruncatedAt *time.Time `json:"truncated_at,omitempty"`
+	Message     *Message   `json:"message,omitempty"`
+}
+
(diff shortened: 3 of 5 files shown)


channel.go
```diff
@@ -184,10 +184,22 @@ func (ch *Channel) Delete() error {
 }
 
 // Truncate removes all messages from the channel.
-func (ch *Channel) Truncate() error {
+// You can pass in options such as hard_delete, skip_push
+// or a custom message.
+func (ch *Channel) Truncate(options ...map[string]interface{}) error {
```
peterdeme, line 189: I used the `...` operator to keep backward compatibility. But now we need to check for the length of the array...

Any better ideas?

ferhatelmas: Define a struct to hold final options and accept a slice option function. Then, initialize final struct with defaults and apply all given option functions.

```go
type truncateOptions struct {
  HardDelete bool
  TruncatedAt time.Time
}

type TruncateOption func(*truncateOptions)

func WithTruncateHardDelete(hardDelete bool) {
  return func(f *finalResult) {
    f.HardDelete = hardDelete
  }
}

func (ch *Channel) Truncate(options ...TruncateOption) error {
  result := &truncateOptions{}
  for _, op := range options {
    op(result)
  }
  // use result as you wish
}

// client
ch.Truncate(
	WithTruncateHardDelete(true),
        WithTruncateAt(oneDayAgo))
```

peterdeme: interesting. I can make it work, although might be a bit complex for such a simple feature :)

CONTRIBUTING.md
```diff
@@ -0,0 +1,26 @@
...
+Contributions to this project are very much welcome, please make sure that your code changes are tested and that follow
+Go best-practices.
+
+## Getting started
+
+### Required environmental variables
+
+The tests require at least two environmental variables: `STREAM_CHAT_API_KEY` and `STREAM_CHAT_API_SECRET`. There are multiple ways to provide that:
+- simply set it in your current shell (`export STREAM_CHAT_API_KEY=xyz`)
+- you could use [direnv](https://direnv.net/)
```
gumuz, line 12: didn't know about this 👌🏼

README.md
```diff
@@ -67,4 +67,4 @@ func main() {
 ### Contributing
 
 Contributions to this project are very much welcome, please make sure that your code changes are tested and that follow
```
gumuz, line 69:
```suggestion
Contributions to this project are very much welcome, please make sure that your code changes are tested and that they follow
```

same nit-picking

channel.go
```diff
@@ -183,11 +183,56 @@ func (ch *Channel) Delete() error {
...
+}
+
 // Truncate removes all messages from the channel.
-func (ch *Channel) Truncate() error {
+// You can pass in options such as hard_delete, skip_push
+// or a custom message.
+func (ch *Channel) Truncate(options ...TruncateOption) error {
+	// TODO: this thingy looks nice, but it's just a workaround to keep
+	// backward compatibility. Whenever we bump a major version, let's just
+	// modify this to: func (ch *Channel) Truncate(options *TruncateOptions)
```
gumuz, line 225: Not sure if I necessarily agree with this, but let's discuss it when we do :)

gumuz (approved):
I left some nit-picking comments, but otherwise looks great 👏🏼

channel.go
peterdeme: I'm open to anything of course. :) I just like to keep things simple.

gumuz (approved):

channel.go
ferhatelmas: I think it's fine as it is. Client is nicer

ferhatelmas: that's why good to drop the comment

ferhatelmas (approved):
LGTM 

(except the comment for todo, not a blocker for merge and/or release)

channel.go
```diff
@@ -183,11 +183,56 @@ func (ch *Channel) Delete() error {
...
+	}
+}
+
+func WithTruncateMessage(message *Message) func(*truncateOptions) {
+	return func(o *truncateOptions) {
+		o.Message = message
+	}
+}
+
+func WithTruncateTruncatedAt(truncatedAt *time.Time) func(*truncateOptions) {
```
yaziine, line 213: Can we rename option functions to: `TruncateWith*` seems more approriate.
Instead of `WithTruncateTruncatedAt` it will be `TruncateWithTruncatedAt`

channel.go
yaziine: And if you don't need any option you can just call it `Truncate()` instead of `Truncate(nil)`

peterdeme: dropped comment 👍

channel.go
peterdeme: renamed

yaziine (approved):
LGTM 🚀

channel.go
ferhatelmas: nice suggestion

PR #152, opened 2021-12-14
feat: add permissions v2

peterdeme:
Permissions V2 endpoints

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -276,6 +276,30 @@ func (ch *Channel) RemoveMembers(ctx context.Context, userIDs []string, message 
 	return nil
 }
 
+type RoleAssignment struct {
+	// UserID is the ID of the user to assign the role to.
+	UserID string `json:"user_id"`
+
+	// ChannelRole is the role to assign to the user.
+	ChannelRole string `json:"channel_role"`
+}
+
+// AssignRoles assigns roles to members with given IDs.
+func (ch *Channel) AssignRole(ctx context.Context, assignments []*RoleAssignment, msg *Message) error {
+	if len(assignments) == 0 {
+		return errors.New("assignments are empty")
+	}
+
+	data := map[string]interface{}{"assign_roles": assignments}
+	if msg != nil {
+		data["message"] = msg
+	}
+
+	p := path.Join("channels", url.PathEscape(ch.Type), url.PathEscape(ch.ID))
+
+	return ch.client.makeRequest(ctx, http.MethodPost, p, nil, data, nil)
+}
+
 type queryMembersResponse struct {
 	Members []*ChannelMember `json:"members"`
 }
--- a/channel_test.go
+++ b/channel_test.go
@@ -95,6 +95,26 @@ func TestChannel_AddMembers(t *testing.T) {
 	assert.Equal(t, user.ID, ch.Members[0].User.ID, "members contain user id")
 }
 
+func TestChannel_AssignRoles(t *testing.T) {
+	c := initClient(t)
+	ctx := context.Background()
+
+	owner := randomUser(t, c)
+	chanID := randomString(12)
+
+	ch, err := c.CreateChannel(ctx, "messaging", chanID, owner.ID, map[string]interface{}{
+		"members": []string{owner.ID},
+	})
+	require.NoError(t, err, "create channel")
+	defer func() {
+		_ = ch.Delete(ctx)
+	}()
+
+	a := []*RoleAssignment{{ChannelRole: "channel_moderator", UserID: owner.ID}}
+	err = ch.AssignRole(ctx, a, nil)
+	require.NoError(t, err)
+}
+
 func TestChannel_QueryMembers(t *testing.T) {
 	c := initClient(t)
 
--- a/channel_type.go
+++ b/channel_type.go
@@ -35,7 +35,7 @@ type (
 	modBehaviour string
 )
 
-type Permission struct {
+type ChannelTypePermission struct {
 	Name   string `json:"name"`   // required
 	Action string `json:"action"` // one of: Deny Allow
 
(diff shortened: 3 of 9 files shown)


permissions.go
```diff
@@ -0,0 +1,78 @@
+package stream_chat //nolint: golint
+
+import (
+	"net/http"
+	"path"
+	"time"
+)
+
+type CustomPermission struct {
```
peterdeme, line 9: I don't like the name but Permission is already used by something else....

permissions.go
```diff
@@ -0,0 +1,78 @@
...
+	UpdatedAt time.Time `json:"updated_at"`
+}
+
+type roleResponse struct {
+	Roles []*Role `json:"roles"`
+}
+
+type Permissions struct {
+	client *Client
+}
```
peterdeme, line 43: I don't like the name `Permissions` in a plural form but `Permission` is already used....

permissions.go
yaziine: `PermissionV2`.

channel.go
```diff
@@ -281,6 +281,26 @@ func (ch *Channel) RemoveMembers(userIDs []string, message *Message) error {
...
+	// UserID is the ID of the user to assign the role to.
+	UserID string `json:"user_id"`
+
+	// ChannelRole is the role to assign to the user.
+	ChannelRole string `json:"channel_role"`
+}
+
+// AssignRoles assigns roles to members with given IDs.
+func (ch *Channel) AssignRole(assignments []*RoleAssignment, msg *Message) error {
+	data := map[string]interface{}{"assign_roles": assignments}
```
yaziine, line 294: can we check `if assignements == nil` first?

permissions.go
peterdeme: good idea

channel.go
peterdeme: ✅

channel_test.go
```diff
@@ -95,6 +95,25 @@ func TestChannel_AddMembers(t *testing.T) {
...
+	owner := randomUser(t, c)
+	chanID := randomString(12)
+
+	ch, err := c.CreateChannel("messaging", chanID, owner.ID, map[string]interface{}{
+		"members": []string{owner.ID},
+	})
+	require.NoError(t, err, "create channel")
+	defer func() {
+		_ = ch.Delete()
+	}()
```
yaziine, line 110: You can use [t.Cleanup](https://pkg.go.dev/testing#T.Cleanup) for this kind of thing.

permissions.go
yaziine: `PermissionClient` ?

yaziine: Actually it makes me think we could probably move each entity to its own package.

permissions.go
```diff
@@ -0,0 +1,78 @@
...
+
+// CreateRole creates a new role.
+func (p *Permissions) CreateRole(name string) error {
+	return p.client.makeRequest(http.MethodPost, "roles", nil, map[string]interface{}{
+		"name": name,
+	}, nil)
+}
+
+// DeleteRole deletes an existing role.
+func (p *Permissions) DeleteRole(name string) error {
```
yaziine, line 53: Same here.

permissions.go
```diff
@@ -0,0 +1,78 @@
...
+
+// ListRole returns all roles.
+func (p *Permissions) ListRoles() ([]*Role, error) {
+	var r roleResponse
+	err := p.client.makeRequest(http.MethodGet, "roles", nil, nil, &r)
+	return r.Roles, err
+}
+
+// GetPermission returns a permission by name.
+func (p *Permissions) GetPermission(id string) (*PermissionV2, error) {
```
yaziine, line 66: Same here.

channel_test.go
peterdeme: Yes, but I wanted to have it consistent with the rest of the tests here

permissions.go
peterdeme: ✅

yaziine:
@peterdeme can you add support to context for Permissions related method please? cf https://github.com/GetStream/stream-chat-go/pull/153

channel_type.go
```diff
@@ -35,7 +35,7 @@ type (
 	modBehaviour string
 )
 
-type Permission struct {
+type ChannelTypePermission struct {
```
yaziine, line 38: Let's rename it to `Policy`  instead.

peterdeme: @yaziine Policy is already used in this namespace
https://github.com/GetStream/stream-chat-go/blob/865e1919bb02070bf6373ea9bba0e30e37bd89ee/app.go#L77

channel_type.go
```diff
@@ -48,8 +48,11 @@ type Permission struct {
 type ChannelType struct {
 	ChannelConfig
 
-	Commands    []*Command    `json:"commands"`
-	Permissions []*Permission `json:"permissions"`
+	Commands []*Command `json:"commands"`
+	// Deprecated: Use Permissions V2 API instead,
+	// that can be found in `permission_client.go.
```
yaziine, line 53:
```suggestion
	// that can be found in permission_client.go.
```

permission_client.go
```diff
@@ -0,0 +1,117 @@
...
+
+// ListRole returns all roles.
+func (p *PermissionClient) ListRoles(ctx context.Context) ([]*Role, error) {
+	var r roleResponse
+	err := p.client.makeRequest(ctx, http.MethodGet, "roles", nil, nil, &r)
+	return r.Roles, err
+}
+
+// CreatePermission creates a new permission.
+func (p *PermissionClient) CreatePermission(ctx context.Context, perm *Permission) error {
```
yaziine, line 76: Can we do some validation on the Permission first?

yaziine: I am wondering which fields are actually mandatory

peterdeme: I would leave that kind of validation to the backend. We never validate such input values anywhere in any SDK.

channel_type.go
ferhatelmas: Actually we could get rid of this and use `Policy` directly because timestamps will be ignored by server to create and data shape is same

peterdeme: Unfortunately `Action` seems to have different type in these classes 🤔

ferhatelmas: let's address in a new PR, we need couple of helpers use these types

ferhatelmas (approved):
LGTM

yaziine (approved):

PR #154, opened 2021-12-14
Expose API error

yaziine:


http.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+func (c *Client) parseResponse(resp *http.Response, result interface{}) error {
+	if resp.Body != nil {
+		defer resp.Body.Close()
+	}
+
+	if resp.StatusCode >= 399 {
+		var apiErr Error
+		err := json.NewDecoder(resp.Body).Decode(&apiErr)
```
thesyncim, line 70: if we fail to decode the body as JSON, we should assign the response body to the apiErr.Message.
in case of IP rate limit we currently don't return the body as json

ferhatelmas: we should do eventually though

http.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+func (c *Client) parseResponse(resp *http.Response, result interface{}) error {
+	if resp.Body == nil {
+		return errors.New("http body is nil")
+	}
+
+	b, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		panic(err)
```
thesyncim, line 69: let's not panic, return errors.New("failed to read HTTP response")

http.go
```diff
@@ -0,0 +1,172 @@
...
+		err := json.Unmarshal(b, &apiErr)
+		if err != nil {
+			// IP rate limit errors sent by our Edge infrastructure are not JSON encoded.
+			// If decode fails here, we need to handle this manually.
+			apiErr.Message = string(b)
+			apiErr.StatusCode = resp.StatusCode
+			return apiErr
+		}
+
+		if resp.StatusCode == http.StatusTooManyRequests {
```
ferhatelmas, line 84: we have this information in every response so every error, it might help somewhere

http.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	if result != nil {
+		return json.Unmarshal(b, result)
+	}
+	return nil
+}
+
+func (c *Client) requestURL(path string, values url.Values) (string, error) {
+	_url, err := url.Parse(c.BaseURL + "/" + path)
```
ferhatelmas, line 97: this pattern is for package globals, better to name without underscore

http_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Millisecond)
+	defer cancel()
+
+	_, err := c.DeleteUsers(ctx, []string{user.ID}, DeleteUserOptions{
+		User:     SoftDelete,
+		Messages: HardDelete,
+	})
+	require.Error(t, err)
+	require.True(t, errors.Is(err, context.DeadlineExceeded))
```
ferhatelmas, line 53: `require.ErrorIs`

http.go
ferhatelmas: good to wrap original error

ferhatelmas (approved):
LGTM

PR #156, opened 2021-12-16
Add base response with rate limit info

yaziine:
This PR adds a base response type `Response` which holds rate limit information computed from response headers.

Breaking changes:
- All methods return now two parameter: a `*Response` and an error
- Except the ones that return specific types, those types have now a `Response` embedded
- Some types have been now exposed eg. `sendFileResponse` is now `SendFileResponse` and contain the `Response`

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -97,6 +97,8 @@ type queryResponse struct {
 	Messages []*Message       `json:"messages,omitempty"`
 	Members  []*ChannelMember `json:"members,omitempty"`
 	Read     []*ChannelRead   `json:"read,omitempty"`
+
+	Response
 }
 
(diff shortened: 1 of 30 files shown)


channel.go
```diff
@@ -118,12 +120,8 @@ func (q queryResponse) updateChannel(ch *Channel) {
 	}
 }
 
-type ImportChannelMessagesResponse struct {
-	Messages []Message `json:"messages"`
-}
-
```
ferhatelmas, line 124: 👍🏻

ferhatelmas (approved):
LGTM

ferhatelmas:
CI is due to unflag, server isn't rolled out yet

driver-devel (approved):

PR #160, opened 2021-12-17
Improve API

yaziine:


ban.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+}
+
+func banWithShadow() func(*banOptions) {
+	return func(opt *banOptions) {
+		opt.Shadow = true
+	}
+}
+
+func banFromChannel(id, _type string) func(*banOptions) {
```
ferhatelmas, line 158: better to swap since we generally use first type then id, not very intuitive

channel.go
```diff
@@ -236,28 +236,45 @@ func (ch *Channel) Truncate(ctx context.Context, options ...TruncateOption) (*Re
 	return &resp, err
 }
 
+type addMembersOptions struct {
```
ferhatelmas, line 239: let's add `assign_roles` support here

ferhatelmas (commented):
Nice touch

ferhatelmas (approved):
LGTM

ferhatelmas:
We can merge it early and find any further improvements/mistakes (if exists)

PR #166, opened 2022-01-05
feat: add grants to channeltype

peterdeme:
In order to fill up the code snippet examples in the [Permission V2 docs page](https://getstream.io/chat/docs/java/user_permissions/?preview=1&language=go), we need to be able to update grants on channel type level.

So basically this is support for ChannelType grants.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -32,6 +32,3 @@ jobs:
           STREAM_CHAT_API_KEY: ${{ secrets.STREAM_CHAT_API_KEY }}
           STREAM_CHAT_API_SECRET: ${{ secrets.STREAM_CHAT_API_SECRET }}
         run: go test -v -race ./...
-
-      - name: Build on ${{ matrix.goVer }}
-        run: go build ./...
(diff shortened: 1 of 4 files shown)


.github/workflows/go.yml
```diff
@@ -32,6 +32,3 @@ jobs:
           STREAM_CHAT_API_KEY: ${{ secrets.STREAM_CHAT_API_KEY }}
           STREAM_CHAT_API_SECRET: ${{ secrets.STREAM_CHAT_API_SECRET }}
         run: go test -v -race ./...
-
-      - name: Build on ${{ matrix.goVer }}
-        run: go build ./...
```
peterdeme, line 37: `go test` already builds a project so this build step was redundant.

ferhatelmas: it might be useful sometimes if there is no test in the package, it will be ignored. In this case, it's flat, good call

ferhatelmas (approved):
LGTM

yaziine (approved):

PR #167, opened 2022-01-12
ci: add release flow

peterdeme:
Here's how an example looks like: https://github.com/GetStream/stream-chat-net/pull/89

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-* @mahboubii @yaziine
+* @ferhatelmas @yaziine
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -11,7 +11,7 @@ concurrency:
 
 jobs:
   test-build:
-    name: Test & Build
+    name: 👷 Test & Build
     runs-on: ubuntu-latest
     strategy:
       max-parallel: 1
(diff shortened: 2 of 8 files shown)


yaziine:
Is it possible to have the changelog directly in the description? We could use something like this: https://github.com/mikepenz/release-changelog-builder-action

peterdeme:
@yaziine it's automatic: https://github.com/GetStream/stream-chat-go/pull/167/files#diff-b93b84f6a7bc16fd654fc88ad84bd92fa34871e6f1b36926523efa7a5fa60ae3R22

The only caveat is that you need to follow [conventional commits](https://www.conventionalcommits.org/) for your commit messages.

ferhatelmas (approved):
LGTM

ferhatelmas:
updated actions and could you modify owners? maybe replace me with Amin

yaziine:
>it's automatic: https://github.com/GetStream/stream-chat-go/pull/167/files#diff-b93b84f6a7bc16fd654fc88ad84bd92fa34871e6f1b36926523efa7a5fa60ae3R22

By manually, I meant the script `get_changelog_diff.js`

yaziine (approved):

ferhatelmas (approved):

ferhatelmas:
> > it's automatic: #167 (files)

> By manually, I meant the script get_changelog_diff.js

if we can get rid of it, definitely better but good for now.

PR #169, opened 2022-01-24
Improve API

yaziine:
#160 was reverted, this PR adds changes back.

### DIFF
--- a/ban_test.go
+++ b/ban_test.go
@@ -1,0 +1,148 @@
+package stream_chat
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestShadowBanUser(t *testing.T) {
+	c := initClient(t)
+	userA := randomUser(t, c)
+	userB := randomUser(t, c)
+	userC := randomUser(t, c)
+
+	ch := initChannel(t, c, userA.ID, userB.ID, userC.ID)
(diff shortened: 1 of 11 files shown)


ban_test.go
```diff
@@ -0,0 +1,148 @@
...
+	})
+	require.NoError(t, err)
+	require.Empty(t, resp.Bans)
+}
+
+func ExampleClient_BanUser() {
+	client, _ := NewClient("XXXX", "XXXX")
+
+	// ban a user for 60 minutes from all channel
+	_, _ = client.BanUser(context.Background(), "eviluser", "modUser", BanWithExpiration(60), BanWithReason("Banned for one hour"))
```
peterdeme, line 136: Shouldn't we make sure there are no errors thrown?

yaziine: Those are examples actually, not test per se. And even if someone runs them from browser, they will fail because there is not a valid client. It's just to give an idea to developer how to use that method.

peterdeme: ohh so that's what `ExampleClient` does... I didn't know we had that

peterdeme (dismissed):
looks good, but can you rename the commit with `feat:` prefix?

ban_test.go
yaziine: https://go.dev/blog/examples

peterdeme: 🤯

peterdeme (approved):

PR #174, opened 2022-02-03
feat: add offset and limit to listimports

peterdeme:
- Add Offset and Limit to import endpoints

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       max-parallel: 1
       matrix:
-        goVer: ["^1.16", "^1.17"]
+        goVer: ["1.16", "1.17"]
     steps:
       - name: Set up Go ${{ matrix.goVer }}
         uses: actions/setup-go@v2
--- a/import.go
+++ b/import.go
@@ -3,6 +3,8 @@ package stream_chat
 import (
 	"context"
 	"net/http"
+	"net/url"
+	"strconv"
 	"time"
 )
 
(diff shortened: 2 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -16,7 +16,7 @@ jobs:
     strategy:
       max-parallel: 1
       matrix:
-        goVer: ["^1.16", "^1.17"]
+        goVer: ["1.16", "1.17"]
```
peterdeme, line 19: image

ferhatelmas: yes, it uses latest by default, previous change was unnecessary

import.go
```diff
@@ -65,9 +72,15 @@ func (c *Client) GetImport(ctx context.Context, id string) (*GetImportResponse,
 // Note: Do not use this.
 // It is present for internal usage only.
 // This function can, and will, break and/or be removed at any point in time.
-func (c *Client) ListImports(ctx context.Context) (*ListImportsResponse, error) {
+func (c *Client) ListImports(ctx context.Context, opts *ListImportsOptions) (*ListImportsResponse, error) {
```
ferhatelmas, line 75: it doesn't matter here but in general, we should have options pattern as a habit/consistency

`ListImports(ctx, WithLimit(10)) // offset is zero` 
`ListImports(ctx, WithLimit(20), WithOffset(10))`

ferhatelmas (approved):

ferhatelmas:
ci is updated :)

.github/workflows/ci.yml
peterdeme: yup, my bad.

can you remove these checks?
image

PR #178, opened 2022-02-07
feat: add quoted message id top options

peterdeme:
Add quoted message id to msg options


message_test.go
```diff
@@ -21,6 +21,12 @@ func TestClient_PinMessage(t *testing.T) {
 	messageResp, err := resp1.Channel.SendMessage(context.Background(), msg, userB.ID)
 	require.NoError(t, err)
 
+	msgWithOptions := &Message{Text: "test message"}
+	_, err = resp1.Channel.SendMessage(context.Background(), msgWithOptions, userB.ID, func(msg *messageRequest) {
```
yaziine, line 25: Can we check that the message is correcty quoted in the response?

peterdeme: ✅

channel_test.go
```diff
@@ -21,6 +21,23 @@ func (ch *Channel) refresh(ctx context.Context) error {
...
 }
 
+func TestClient_TestQuery(t *testing.T) {
+	c := initClient(t)
+	membersID := randomUsersID(t, c, 1)
+	ch := initChannel(t, c, membersID...)
+	ctx := context.Background()
+	_, err := ch.SendMessage(ctx, &Message{Text: "test message", Pinned: true}, ch.CreatedBy.ID)
+	require.NoError(t, err)
+	defer func() {
```
yaziine, line 31: nit: since go1.14, we have [t.Cleanup()](https://pkg.go.dev/testing#T.Cleanup) which does this for you.

yaziine (approved):

channel_test.go
peterdeme: I'll try to unify it in the future for every test method.

PR #180, opened 2022-02-14
feat: add all app settings

peterdeme:


peterdeme:
We should refactor this with some inheritance logic with the next major...

app.go
```diff
@@ -7,19 +7,30 @@ import (
...
-	AutoTranslationEnabled *bool               `json:"auto_translation_enabled,omitempty"`
-	Grants                 map[string][]string `json:"grants,omitempty"`
-	MigratePermissionsToV2 *bool               `json:"migrate_permissions_to_v2,omitempty"`
-	PermissionVersion      string              `json:"permission_version,omitempty"`
-	FileUploadConfig       *FileUploadConfig   `json:"file_upload_config,omitempty"`
-	ImageUploadConfig      *FileUploadConfig   `json:"image_upload_config,omitempty"`
+	DisableAuth               *bool               `json:"disable_auth_checks,omitempty"`
+	DisablePermissions        *bool               `json:"disable_permissions_checks,omitempty"`
+	APNConfig                 *APNConfig          `json:"apn_config,omitempty"`
+	FirebaseConfig            *FirebaseConfig     `json:"firebase_config,omitempty"`
```
ferhatelmas, line 13: while we're here, we could add huawei/xiaomi configs too

app.go
```diff
@@ -37,13 +51,13 @@ func (a *AppSettings) SetAPNConfig(c APNConfig) *AppSettings {
 	return a
 }
 
-func (a *AppSettings) SetFirebaseConfig(c FirebaseConfig) *AppSettings {
+func (a *AppSettings) SetFirebaseConfig(c FirebaseConfigRequest) *AppSettings {
```
ferhatelmas, line 54: this separation is a good direction but beware that this is a breaking change

peterdeme: Should we raise major? I don’t know how many people uses this

ferhatelmas (approved):

app.go
ferhatelmas: I am fine with big major number as much as we're semantic

PR #182, opened 2022-02-16
docs: unify READMEs across backend repos

peterdeme:
Unify READMEs across backend repos. looks like this: https://github.com/GetStream/stream-chat-go/blob/readme-v2/README.md


ferhatelmas (approved):
LGTM

Issue #183, opened 2022-02-24
Channel Query function does not expose options


kimmom:
[Stream docs Go example](https://getstream.io/chat/docs/go-golang/channel_pagination/?language=go) uses function ch.query which is not exposed. So the example does not work.

```go
options := map[string]interface{}{ 
	"messages": map[string]interface{}{"limit": 50, "id_lt": lastMessageId}, 
	"members":  map[string]interface{}{"limit": 10, "offset": 0}, 
	"watchers": map[string]interface{}{"limit": 20, "offset": 0}, 
} 
 
ch.query(ctx, options, nil)
```

Changing the function to exposed ch.Query doesn't work either since that function does not allow passing options to it.

Thus I found no way to query channel messages with pagination using the Go client.

ferhatelmas:
Yes, it makes sense to export it and we will do it and update the docs accordingly

cc @peterdeme

peterdeme:
Fix in https://github.com/GetStream/stream-chat-go/pull/184

ferhatelmas:
https://github.com/GetStream/stream-chat-go/releases/tag/v5.1.0

PR #184, opened 2022-02-25
feat: add query with options

peterdeme:


peterdeme:
This solves https://github.com/GetStream/stream-chat-go/issues/183

ferhatelmas (approved):
Good as it is but as a next iteration, it would be better to have options a specific type instead of a map

PR #193, opened 2022-03-04
feat: add CreateImportURL method

gumuz:
added new private method CreateImportURL for internal use only

### DIFF
--- a/import.go
+++ b/import.go
@@ -16,7 +16,7 @@ type ImportTaskHistory struct {
 
 type ImportTask struct {
 	CreatedAt time.Time            `json:"created_at"`
-	Filename  string               `json:"filename"`
+	Path      string               `json:"path"`
 	History   []*ImportTaskHistory `json:"history"`
 	ID        string               `json:"id"`
 	State     string               `json:"state"`
@@ -32,7 +32,12 @@ type ListImportsOptions struct {
 
 type CreateImportResponse struct {
 	ImportTask *ImportTask `json:"import_task"`
-	UploadURL  string      `json:"upload_url"`
+	Response
+}
+
+type CreateImportURLResponse struct {
+	Path      string `json:"path"`
+	UploadURL string `json:"upload_url"`
 	Response
 }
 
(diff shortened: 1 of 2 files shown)


peterdeme (approved):
lgtm

import.go
```diff
@@ -46,13 +51,24 @@ type ListImportsResponse struct {
 	Response
 }
 
+// CreateImportURL creates a new import URL.
+// Note: Do not use this.
+// It is present for internal usage only.
+// This function can, and will, break and/or be removed at any point in time.
+func (c *Client) CreateImportURL(ctx context.Context, filename string) (*CreateImportURLResponse, error) {
+	var resp CreateImportURLResponse
+	err := c.makeRequest(ctx, http.MethodPost, "imports_url", nil, map[string]string{"filename": filename}, &resp)
```
ferhatelmas, line 60: for the correct grammar, it sounds like `import_urls`

import.go
```diff
@@ -46,13 +51,24 @@ type ListImportsResponse struct {
...
+
+	return &resp, err
+}
+
 // CreateImport creates a new import task.
 // Note: Do not use this.
 // It is present for internal usage only.
 // This function can, and will, break and/or be removed at any point in time.
-func (c *Client) CreateImport(ctx context.Context, filename string) (*CreateImportResponse, error) {
+func (c *Client) CreateImport(ctx context.Context, path string) (*CreateImportResponse, error) {
```
ferhatelmas, line 69: `path` standard package name, might be better to have a different name

peterdeme: surprised that the linter doesn't catch this 🤔

ferhatelmas: because it's not imported in that file at the moment but it could be later on

gumuz: ok, let me think, 'cause filepath is also taken already 😅

yaziine: `fp` 😉

yaziine: also you need to change in `CreateImportURL` #consistency

peterdeme (approved):

ferhatelmas (approved):

PR #196, opened 2022-03-09
fix: check if iat is zero

peterdeme:
- Check if iat value is zero


ferhatelmas (approved):

client.go
```diff
@@ -127,7 +127,7 @@ func (c *Client) CreateToken(userID string, expire time.Time, issuedAt ...time.T
 	if !expire.IsZero() {
 		claims["exp"] = expire.Unix()
 	}
-	if len(issuedAt) > 0 {
```
ferhatelmas, line 130: how about a tiny test?

peterdeme: ✅

PR #212, opened 2022-03-31
feat(truncate): add user id if message provided

peterdeme:
Add truncater person


channel.go
```diff
@@ -263,9 +267,16 @@ func TruncateWithSkipPush() func(*truncateOptions) {
 	}
 }
 
-func TruncateWithMessage(message *Message) func(*truncateOptions) {
+func TruncateWithMessage(message *Message, userID string) func(*truncateOptions) {
```
peterdeme, line 270: this is a breaking change but I assume this is a breaking change in the backend side as well: https://github.com/GetStream/chat/blob/master/controllers/truncate_channel.go#L77=

ferhatelmas: this isn't needed since `user_id` is passed inside `message` while the other is needed since we can truncate without message

channel_test.go
```diff
@@ -404,7 +404,7 @@ func TestChannel_TruncateWithOptions(t *testing.T) {
 	// Now truncate it
 	_, err = ch.Truncate(ctx,
 		TruncateWithSkipPush(),
-		TruncateWithMessage(&Message{Text: "truncated channel", User: &User{ID: user.ID}}),
+		TruncateWithMessage(&Message{Text: "truncated channel", User: &User{ID: user.ID}}, user.ID),
```
ferhatelmas, line 407: https://github.com/GetStream/stream-chat-go/pull/212/files#r841792025=

you can see it here in action via `User` in message. Let's add other option here too

channel.go
peterdeme: okay, I'll remove this

channel_test.go
ferhatelmas: can we use the new option and see if it's set in the response?

peterdeme: ✅

channel_test.go
```diff
@@ -402,14 +402,16 @@ func TestChannel_TruncateWithOptions(t *testing.T) {
 	assert.Equal(t, ch.Messages[0].ID, resp.Message.ID, "message exists")
 
 	// Now truncate it
-	_, err = ch.Truncate(ctx,
+	truncatResp, err := ch.Truncate(ctx,
 		TruncateWithSkipPush(),
 		TruncateWithMessage(&Message{Text: "truncated channel", User: &User{ID: user.ID}}),
+		TruncateWithUser(&User{ID: user.ID}),
```
ferhatelmas, line 408: let's set another user here then compare message and truncator are different

channel.go
```diff
@@ -269,16 +274,34 @@ func TruncateWithMessage(message *Message) func(*truncateOptions) {
...
+	Response
+	Channel *Channel `json:"channel"`
+	Message *Message `json:"message"`
+}
+
 // Truncate removes all messages from the channel.
 // You can pass in options such as hard_delete, skip_push
 // or a custom message.
-func (ch *Channel) Truncate(ctx context.Context, options ...TruncateOption) (*Response, error) {
+func (ch *Channel) Truncate(ctx context.Context, options ...TruncateOption) (*TruncateResponse, error) {
```
ferhatelmas, line 304: this is better response but breaking 🤦🏻‍♂️

peterdeme: ✅

ferhatelmas (approved):
LGTM

PR #223, opened 2022-05-18
feat(campaign): add campaign endpoints

peterdeme:


campaign.go
```diff
@@ -0,0 +1,144 @@
...
+
+type CampaignData struct {
+	SegmentID         string            `json:"segment_id,omitempty"`
+	SenderID          string            `json:"sender_id,omitempty"`
+	ChannelType       string            `json:"channel_type,omitempty"`
+	Text              string            `json:"text,omitempty"`
+	Defaults          map[string]string `json:"defaults,omitempty"`
+	Filter            SegmentFilter     `json:"filter,omitempty"`
+	Attachments       []*Attachment     `json:"attachments,omitempty"`
+	PushNotifications bool              `json:"push_notifications,omitempty"`
```
yaziine, line 36: PushNotifications will be removed, you can assume this and remove it

yaziine: actually, we are actively working on campaigns, i suspect changes may come in the future (not 100% sure but there is a risk)), i think better to postpone this PR a bit until we are 100% sure on API contract. @ferhatelmas what do you think?

ferhatelmas: sg, good to handle untyped languages first

peterdeme: 👍 those are done.
closing the issue.

peterdeme:
Will be done once finalized

PR #228, opened 2022-07-11
Handle clearing in settings

### LINKED ISSUE #227: Remove SQS config
I used the library to add my SQS config on my app.
But I can remove it. If I try to set it as empty string it still doesn't work.

Can you add a way to do it?

ferhatelmas:
fixes #227


yaziine (approved):
Thank you @ferhatelmas

PR #257, opened 2023-11-05
refactor(app): add SetEnforceUniqueUsernames to restrict allowed values

jnelle:
## CLA

- [X] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required).
- [X] The code changes follow best practices
- [X] Code changes are tested (add some information if not applicable)

Since I couldn't find the allowed values in the GO SDK docs, I decided to add the SetEnforceUniqueUsernames function, because it restricts the allowed values and shows the developer which values are allowed. If neccessary I could write a test

### DIFF
--- a/app.go
+++ b/app.go
@@ -54,6 +54,36 @@ type AppSettings struct {
 	AsyncModerationConfig     *AsyncModerationConfiguration `json:"async_moderation_config,omitempty"`
 }
 
+type EnforceUniqueUsernames int64
+
+const (
+	No EnforceUniqueUsernames = iota
+	App
+	Team
+)
+
+// check allowed values for enforce_unique_usernames here
+// https://getstream.io/chat/docs/rest/#settings-updateapp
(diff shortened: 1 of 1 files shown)


app.go
```diff
@@ -54,6 +54,36 @@ type AppSettings struct {
...
+	case No:
+		result = "no"
+	case App:
+		result = "app"
+	case Team:
+		result = "team"
+	default:
+		result = "app"
+	}
+	return &result
```
dudemelo, line 79: Does this need to be a pointer? Because I believe that Go's "escape analysis" will put this in the heap instead of stack.

This beauty thing Go does to avoid dangling pointer—since variable should only exists inside the function—always bug my brain. :exploding_head:

jnelle: You're right, it doesn't have to be. Gonna change it later

PR #277, opened 2024-05-22
Modernize

guerinoni:
- chore: update golangci-linter
- chore: drop support go 1.17
- ci: bump GH actions
- fix: format code with gofumpt


akupila (dismissed):
🚀

akupila:
CI was waiting for `1.17` to run. I removed that branch protection rule so it's no longer needed (as it would never run)

PR #293, opened 2024-12-04
[CHA-14] Added support for pinning, archiving channels and partial member update

totalimmersion:
https://linear.app/stream/issue/CHA-14/members-custom-data-go-sdk

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -898,3 +898,39 @@ func (ch *Channel) Unpin(ctx context.Context, userID string) (*ChannelMemberResp
 	err := ch.client.makeRequest(ctx, http.MethodPatch, p, nil, data, resp)
 	return resp, err
 }
+
+func (ch *Channel) Archive(ctx context.Context, userID string) (*ChannelMemberResponse, error) {
+	if userID == "" {
+		return nil, errors.New("user ID must be not empty")
+	}
+
+	p := path.Join("channels", url.PathEscape(ch.Type), url.PathEscape(ch.ID), "member", url.PathEscape(userID))
+
+	data := map[string]interface{}{
+		"set": map[string]interface{}{
+			"archived": true,
+		},
+	}
+
+	resp := &ChannelMemberResponse{}
+	err := ch.client.makeRequest(ctx, http.MethodPatch, p, nil, data, resp)
+	return resp, err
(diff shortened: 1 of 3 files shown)


channel.go
```diff
@@ -27,6 +27,8 @@ type ChannelMember struct {
 	InviteAcceptedAt *time.Time `json:"invite_accepted_at,omitempty"`
 	InviteRejectedAt *time.Time `json:"invite_rejected_at,omitempty"`
 	Role             string     `json:"role,omitempty"`
+	ArchivedAt       *time.Time `json:"archived_at,omitempty"`
```
akupila, line 30:
```suggestion
```

akupila (commented):
~This only adds support for pinning, so we should probably exclude `ArchivedAt` (or also add that here, as it's very similar)~

channel.go
totalimmersion: it's still work in progress, i'll convert it to draft

channel_test.go
```diff
@@ -647,6 +683,108 @@ func TestChannel_Mute_Unmute(t *testing.T) {
...
+	ctx := context.Background()
+	users := randomUsers(t, c, 5)
+
+	members := make([]string, 0, len(users))
+	for i := range users {
+		members = append(members, users[i].ID)
+	}
+	ch := initChannel(t, c, members...)
+
+	//pin the channel
```
akupila, line 748:
```suggestion
	//archive the channel
```

channel_test.go
```diff
@@ -647,6 +683,108 @@ func TestChannel_Mute_Unmute(t *testing.T) {
...
+	channels := queryChannResp.Channels
+	require.NoError(t, err, "query archived channel")
+	require.Len(t, channels, 1)
+	require.Equal(t, channels[0].CID, ch.CID)
+
+	member, err = ch.Unarchive(ctx, users[0].ID)
+	require.NoError(t, err, "unarchive channel")
+	require.Nil(t, member.ChannelMember.ArchivedAt)
+
+	// query for pinned the channel
```
akupila, line 773:
```suggestion
	// query for the archived channel
```

akupila (dismissed):

PR #336, opened 2025-07-14
[CHA-941]: update channel to accept ChannelMembers too

rafaelmf3:
# Submit a pull request

## CLA

- [ ] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required).
- [ ] The code changes follow best practices

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -421,7 +421,7 @@ func (ch *Channel) AddMembers(ctx context.Context, userIDs []string, options ...
 	}
 
 	opts := &addMembersOptions{
-		MemberIDs: userIDs,
+		ChannelMembers: newChannelMembersFromStrings(userIDs),
 	}
 
 	for _, fn := range options {
--- a/channel.go
+++ b/channel.go
@@ -42,8 +42,8 @@ type ChannelMember struct {
 	UpdatedAt time.Time `json:"updated_at,omitempty"`
 }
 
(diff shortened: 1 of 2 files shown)


channel.go
```diff
@@ -42,6 +42,17 @@ type ChannelMember struct {
...
+// NewChannelMembersFromStrings creates a ChannelMembers from a slice of strings
+func NewChannelMembersFromStrings(members []string) []*ChannelMember {
+	channelMembers := make([]*ChannelMember, len(members))
+	for i, m := range members {
+		channelMembers[i] = &ChannelMember{
+			UserID: m,
+		}
+	}
+	return channelMembers
+}
```
totalimmersion, line 54: Why do we need this function?

rafaelmf3: I wrote it to change the AddMembers and CreateChannel to use ChannelMembers from []string, but you are right it can be removed, the old way work as well

totalimmersion: ok, but wouldn't it be better to use an unexported function and thus make it private? It doesn't look like you want to expose this in the sdk, do you?

rafaelmf3: I made it private

PR #391, opened 2026-05-07
feat(webhooks): verifyAndParse* API for compressed payloads (CHA-3071)

nijeesh-stream:
## Summary

Adds first-class support for **gzip-compressed webhook payloads** (HTTP webhooks, SQS, SNS) and exposes a stable `VerifyAndParse*` API that mirrors the cross-SDK contract published in [Webhooks Overview](https://getstream.io/chat/docs/node/webhooks_overview/#handling-the-webhook).

### New public API

Primitives (in `webhook.go`, intended to be composable):

- `GunzipPayload(body) ([]byte, error)` — gzip-magic-byte detection, no-op when not compressed
- `DecodeSqsPayload(body) ([]byte, error)` — base64 decode then gunzip-if-magic
- `DecodeSnsPayload(notificationBody) ([]byte, error)` — JSON-parse the SNS HTTP notification envelope, extract the inner `Message`, then run the SQS pipeline. Falls through to a pre-extracted `Message` string when the input is not a JSON envelope
- `VerifySignature(body, signature, secret) error` — HMAC-SHA256 over the **uncompressed** body, with a constant-time comparison (matters for the HTTP webhook path where the `X-Signature` header is exposed publicly; SQS / SNS deliveries arrive over AWS-internal transports where timing-attack resistance is not strictly required). Returns a non-nil error wrapping `stream.ErrInvalidWebhook` on mismatch — see Unified error handling below.
- `ParseEvent(payload) (*Event, error)` — JSON → typed `*Event`

Composites (return a typed `*Event`):

- `VerifyAndParseWebhook(body, signature, secret) (*Event, error)`
- `VerifyAndParseSqs(body, signature, secret) (*Event, error)`
- `VerifyAndParseSns(body, signature, secret) (*Event, error)`

Client-bound versions on `*Client` use the configured app secret automatically.

### Backwards compatibility

`Client.VerifyWebhook` is preserved and delegates to `VerifySignature`. Existing callers continue to work unchanged for plain (uncompressed) bodies.

### Unified error handling

Per cross-SDK coordination (mogita's review across the 6 sibling SDK PRs), every webhook failure path now terminates at a **single sentinel error — `stream.ErrInvalidWebhook`** so customers only need one error-handling arm. The message text identifies which failure mode fired so callers that want to differentiate (security logging, retry policy) can filter on substring:

| Failure mode      | Message                       |
| ----------------- | ----------------------------- |
| Signature mismatch | `signature mismatch`         |
| Base64 decode      | `invalid base64 encoding`    |
| Gzip decompression | `gzip decompression failed`  |
| JSON parse         | `invalid JSON payload`       |

Use `errors.Is(err, stream.ErrInvalidWebhook)` for the unified check; the inner cause (when present) is preserved via `%w` wrapping. `VerifySignature` now returns `error` (was `bool`) so it surfaces the mode-specific message; the legacy `Client.VerifyWebhook` (bool return, constant-time `hmac.Equal`) is untouched.

### Tests

`webhook_test.go` covers plain / gzip / base64 / base64+gzip payloads, signature mismatches, malformed bytes, JSON parsing into `*Event`, and a backwards-compat path for `Client.VerifyWebhook`. Linked Linear ticket: [CHA-3071](https://linear.app/stream/issue/CHA-3071/compress-webhook-payloads).

### Golden test fixtures (Tommaso)

Added shared reference fixtures to the test suite so future SDKs can sanity-check decoders against the same payloads:

```
aGVsbG93b3JsZA==                          -> helloworld   (base64)
H4sIAGrYAWoAA8tIzcnJL88vykkBAK0g6/kKAAAA -> helloworld   (base64 + gzip)
```
## Test plan

- [x] `go vet ./...` clean
- [x] `go test -c .` (test binary compiles clean)
- [ ] Full `go test ./...` runs in CI (package `init()` requires `STREAM_KEY` / `STREAM_SECRET`)

### DIFF
--- a/webhook.go
+++ b/webhook.go
@@ -13,10 +13,9 @@ import (
 	"io"
 )
 
-// ErrInvalidWebhookSignature is returned by VerifyAndParse* helpers when
-// the supplied signature does not match the HMAC computed over the
-// uncompressed JSON body.
-var ErrInvalidWebhookSignature = errors.New("invalid webhook signature")
+// ErrInvalidWebhook is the sentinel wrapped by every malformed-webhook /
+// verification error from this package so callers can use errors.Is with a single check.
+var ErrInvalidWebhook = errors.New("invalid webhook")
 
 var gzipMagic = []byte{0x1f, 0x8b}
 
@@ -32,15 +31,15 @@ func UngzipPayload(body []byte) ([]byte, error) {
 	}
 	zr, err := gzip.NewReader(bytes.NewReader(body))
 	if err != nil {
-		return nil, fmt.Errorf("decompress gzip payload: %w", err)
+		return nil, fmt.Errorf("gzip decompression failed: %w", ErrInvalidWebhook)
 	}
 	out, err := io.ReadAll(zr)
 	if err != nil {
 		_ = zr.Close()
-		return nil, fmt.Errorf("read gzip payload: %w", err)
+		return nil, fmt.Errorf("gzip decompression failed: %w", ErrInvalidWebhook)
 	}
 	if err := zr.Close(); err != nil {
-		return nil, fmt.Errorf("finalize gzip payload: %w", err)
+		return nil, fmt.Errorf("gzip decompression failed: %w", ErrInvalidWebhook)
 	}
 	return out, nil
 }
@@ -52,7 +51,7 @@ func UngzipPayload(body []byte) ([]byte, error) {
 func DecodeSqsPayload(body string) ([]byte, error) {
 	decoded, err := base64.StdEncoding.DecodeString(body)
 	if err != nil {
-		return nil, fmt.Errorf("base64-decode payload: %w", err)
+		return nil, fmt.Errorf("invalid base64 encoding: %w", ErrInvalidWebhook)
 	}
 	return UngzipPayload(decoded)
 }
@@ -90,16 +89,22 @@ func extractSnsMessage(notificationBody string) (string, bool) {
 	return *envelope.Message, true
 }
 
-// VerifySignature returns true when signature equals the hex-encoded
-// HMAC-SHA256 of body using secret as the key. The comparison is
-// constant-time. The signature is always computed over the
-// uncompressed JSON bytes, so callers that decoded a gzipped or
-// base64-wrapped payload must pass the inflated bytes here.
-func VerifySignature(body []byte, signature, secret string) bool {
+func signatureMatch(body []byte, signature, secret string) bool {
 	mac := hmac.New(sha256.New, []byte(secret))
 	_, _ = mac.Write(body)
 	expected := []byte(hex.EncodeToString(mac.Sum(nil)))
-	return hmac.Equal(expected, []byte(signature))
+	return len(signature) == len(expected) && hmac.Equal(expected, []byte(signature))
+}
+
+// VerifySignature compares the hex-encoded HMAC-SHA256 of body (using secret as the key)
+// to signature with a constant-time comparison. It returns nil on match.
+// On mismatch it returns an error wrapping ErrInvalidWebhook (message contains
+// "signature mismatch"). The digest is always over uncompressed JSON bytes.
+func VerifySignature(body []byte, signature, secret string) error {
+	if !signatureMatch(body, signature, secret) {
+		return fmt.Errorf("signature mismatch: %w", ErrInvalidWebhook)
+	}
+	return nil
 }
 
 // ParseEvent decodes the JSON-encoded webhook payload into a typed
@@ -108,22 +113,20 @@ func VerifySignature(body []byte, signature, secret string) bool {
 func ParseEvent(payload []byte) (*Event, error) {
 	var ev Event
 	if err := json.Unmarshal(payload, &ev); err != nil {
-		return nil, fmt.Errorf("parse webhook event: %w", err)
+		return nil, fmt.Errorf("invalid JSON payload: %w", ErrInvalidWebhook)
 	}
 	return &ev, nil
 }
 
 func verifyAndParse(payload []byte, signature, secret string) (*Event, error) {
-	if !VerifySignature(payload, signature, secret) {
-		return nil, ErrInvalidWebhookSignature
+	if err := VerifySignature(payload, signature, secret); err != nil {
+		return nil, err
 	}
 	return ParseEvent(payload)
 }
 
 // VerifyAndParseWebhook decompresses body when gzipped, verifies the
-// HMAC signature against secret, and returns the parsed Event. Returns
-// ErrInvalidWebhookSignature on mismatch and a wrapped error on any
-// decode failure.
+// HMAC signature against secret, and returns the parsed Event.
 func VerifyAndParseWebhook(body []byte, signature, secret string) (*Event, error) {
 	inflated, err := UngzipPayload(body)
 	if err != nil {
--- a/webhook.go
+++ b/webhook.go
@@ -20,13 +20,13 @@ var ErrInvalidWebhookSignature = errors.New("invalid webhook signature")
 
 var gzipMagic = []byte{0x1f, 0x8b}
 
-// UngzipPayload returns body unchanged unless the first two bytes are
+// GunzipPayload returns body unchanged unless the first two bytes are
 // the gzip magic (1f 8b, per RFC 1952), in which case the gzip stream
 // is inflated and the decompressed bytes are returned.
 //
 // Magic-byte detection lets the same handler stay correct when
 // middleware auto-decompresses the request before your code sees it.
-func UngzipPayload(body []byte) ([]byte, error) {
+func GunzipPayload(body []byte) ([]byte, error) {
 	if len(body) < 2 || !bytes.Equal(body[:2], gzipMagic) {
 		return body, nil
 	}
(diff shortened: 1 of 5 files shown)


webhook.go
```diff
@@ -0,0 +1,143 @@
...
+	decoded, err := base64.StdEncoding.DecodeString(body)
+	if err != nil {
+		return nil, fmt.Errorf("base64-decode payload: %w", err)
+	}
+	return UngzipPayload(decoded)
+}
+
+// DecodeSnsPayload is byte-for-byte identical to DecodeSqsPayload;
+// exposed under both names so call sites read intent.
+func DecodeSnsPayload(message string) ([]byte, error) {
```
mogita: **SNS envelope unwrapping vs the agreed contract**

This aliases the SQS pipeline (base64 + ungzip), but the spec is `decode_sns_payload(notification_body: string)` — the parameter name implies the full SNS HTTP notification body. Real SNS deliveries are JSON envelopes:

```json
{"Type":"Notification","MessageId":"...","Message":"<base64+gzip payload>","MessageAttributes":{...}}
```

Right now the SDK requires every caller to JSON-parse the envelope and pull out `Message` themselves before calling this. If a user passes the raw HTTP body in, signature verification runs against the envelope JSON, not the payload, and fails silently.

Same gap exists across all 6 SDKs in this rollout (JS/Go/Ruby/PHP/Java/.NET) — needs a contract decision before merge:
- **Option A (preferred):** parse the envelope here, extract `Message`, then run the base64+gzip pipeline. Matches the parameter name. Removes a caller footgun.
- **Option B:** rename to `DecodeSnsMessage` and document explicitly that callers pre-extract.

Please align all SDKs on one path.

webhook_test.go
```diff
@@ -0,0 +1,280 @@
...
+
+	t.Run("invalid base64 surfaced as error", func(t *testing.T) {
+		got, err := c.VerifyAndParseSqs("!!!not-base64!!!", sig)
+		require.Error(t, err)
+		require.False(t, errors.Is(err, ErrInvalidWebhookSignature))
+		require.Nil(t, got)
+	})
+}
+
+func TestVerifyAndParseSns(t *testing.T) {
```
mogita: **Missing real SNS envelope coverage**

`TestVerifyAndParseSns` only exercises an already-extracted `Message` string, never a full SNS HTTP notification envelope (`{"Type":"Notification","MessageId":"...","Message":"<base64+gzip>","MessageAttributes":{...}}`). Whatever the contract decision is on `DecodeSnsPayload` (see the other comment), this gap masks the behavior.

Recommend adding a shared fixture across all 6 SDKs using the same SNS envelope JSON, so behavior is identical and verifiable.

mogita (commented):
Cross-SDK review pass for CHA-3071. Two inline comments — see below.

webhook.go
nijeesh-stream: Addressed in 8db5b2c with Option A: `DecodeSnsPayload` now parses the SNS HTTP notification envelope (`{"Type":"Notification","Message":"..."}`) and extracts the inner `Message` field before running the SQS pipeline. When the input is not a JSON envelope it falls through to the pre-extracted `Message` path so existing callers keep working. Same commit also fixes legacy `Client.VerifyWebhook` to use `hmac.Equal` for constant-time comparison (was `bytes.Equal`). All 7 SDKs aligned on this contract.

webhook_test.go
nijeesh-stream: Added in 8db5b2c — `TestDecodeSnsPayload` now covers `full SNS HTTP notification envelope` and an `envelope with whitespace prefix` case; `TestVerifyAndParseSns` adds a `full SNS HTTP notification envelope` case plus an `envelope signature verifies against inner payload` guard. The `snsEnvelope()` test helper matches the shared shape across all 7 SDKs (Type/MessageId/TopicArn/Message/MessageAttributes/X-Signature).

mogita:
**Cross-SDK coordination: unifying webhook exception types**

After the review pass across all 6 SDKs in this rollout and team discussion, we're consolidating the new webhook exception strategy to a **single unified error type** rather than the split (signature vs parse sentinels) being introduced in this PR.

The [Webhook Handling Spec on Notion (CHA-2961)](https://www.notion.so/stream-wiki/Server-Side-SDK-Webhook-Handling-Spec-34b6a5d7f9f681e78003c443f227493c) has been revised to reflect this — §5.2 / §5.3 / §7 now specify a single class.

**Why unified:** From a customer's perspective, all failure modes — signature mismatch, gzip decompression failure, base64 decode failure, SNS envelope failure, JSON parse failure, missing schema field — terminate at the same error-handling branch in customer code. A signature/parse split adds structural complexity without changing customer behavior. Customers who want to filter security logs for signature mismatches specifically can do so via error message text or `errors.Unwrap` cause-chain.

**Sentinel name:** `ErrInvalidWebhook` (+ struct type `InvalidWebhookError`) — "Invalid" covers all failure modes accurately and is consistent with Go stdlib naming (`errors.ErrUnsupported`, etc., where the prefix `Err` names the category).

Per-SDK naming across the rollout:

| SDK | Class name |
|-----|-----------|
| JS | `InvalidWebhookError` (extends `Error`) |
| Python | `InvalidWebhookError` |
| Go | sentinel `ErrInvalidWebhook` + struct `InvalidWebhookError` |
| Java | `InvalidWebhookException` (extends existing `StreamException`) |
| PHP | `InvalidWebhookException` (extends existing `StreamException`) |
| Ruby | `StreamChat::InvalidWebhookError` (extends `StandardError`) |
| .NET | `StreamInvalidWebhookException` (extends `StreamBaseException`) |

**Asks for this PR:**

1. Consolidate the existing sentinels (`ErrInvalidSignature`, `ErrMalformedWebhook`, etc.) into a single `ErrInvalidWebhook` sentinel, with concrete struct `InvalidWebhookError` carrying details
2. Wrap **all** failure paths to return that single sentinel — signature mismatch, gzip failure, base64 failure, SNS envelope failure, JSON parse failure, missing `type`/schema failure — using `fmt.Errorf("%w: <mode>: %v", ErrInvalidWebhook, cause)` so `errors.Is(err, ErrInvalidWebhook)` matches and `errors.Unwrap` reaches the cause
3. The wrapped message identifies which failure mode fired (`"signature mismatch"`, `"invalid base64"`, `"missing type field"`) so customers can filter on substring
4. Legacy `Client.VerifyWebhook` (returning `bool`) stays unchanged — back-compat preserved. Note: there's an existing review comment on `AppClient.cs`-equivalent in another SDK about the legacy method's `bytes.Equal` not being timing-safe; same applies to `client.go:160`, but that's separate from this exception-unification ask.
5. Update unit tests to assert via `errors.Is(err, stream.ErrInvalidWebhook)`; for mode-specific tests, also assert on error message substrings

This same comment is being posted on all 6 SDK PRs (JS / Go / Ruby / PHP / Java / .NET) for coordination. Happy to discuss naming or scope tradeoffs.

webhook.go
mogita: The [SQS & SNS Payload Compression spec](https://linear.app/stream/document/sqs-and-sns-payload-compression-spec-e95df9020e08) just landed on the Linear ticket. It reaffirms that SNS deliveries arrive as JSON envelopes `{"Type":"Notification","Message":"<base64+gzip>",...}` with the actual payload inside `Message`. SQS-via-SNS-subscribers forwards the same wire format.

The SNS envelope-unwrap question I raised earlier on this line is still the open blocker. Contract resolution needed before any of these PRs merge:

- **(A)** `decode_sns_payload` parses the envelope, extracts `Message`, then base64 + gunzip — hides the envelope shape from customers, single call site.
- **(B)** Rename helper to `decode_sns_message` and document that the caller pre-extracts `Message`.

Pick one and apply uniformly across all 6 SDKs. I'm leaning **(A)** to spare every customer from a JSON unwrap step.

nijeesh-stream: Option A is already live as of `8db5b2c` (yesterday). `DecodeSnsPayload(notificationBody string) ([]byte, error)` now JSON-parses the SNS HTTP notification envelope, extracts `Message`, then runs the SQS pipeline (base64 → gunzip-if-magic). Lenient fallback retained for callers who already extracted `Message` themselves. `VerifyAndParseSns` delegates to it. Tests in `webhook_test.go` cover `full SNS HTTP notification envelope`, an `envelope with whitespace prefix` case, and an `envelope signature verifies against inner payload` guard. Re-resolving.

nijeesh-stream:
Shipped in [`c32c92a`](https://github.com/GetStream/stream-chat-go/pull/391/commits/c32c92a2754e72b549dd4519bc678f8511574247) along with the SNS envelope fix.

- Sentinel-only variant per the maintainer's call (no separate struct type for `errors.As`)
- `ErrInvalidWebhookSignature` → `ErrInvalidWebhook`
- Every primitive wraps with `fmt.Errorf("<mode>: %v: %w", err, ErrInvalidWebhook)` so `errors.Is(err, stream.ErrInvalidWebhook)` is the single check; modes are `signature mismatch`, `invalid base64 encoding`, `gzip decompression failed`, `invalid JSON payload`
- `VerifySignature` now returns `error` (was `bool`) so it surfaces the mode token; legacy `Client.VerifyWebhook` (bool return, constant-time `hmac.Equal`) is untouched
- `webhook_test.go`: three failure-mode subtests now assert both `errors.Is(err, ErrInvalidWebhook)` AND the substring; `go build ./...`, `go vet ./...`, `gofmt -l webhook.go webhook_test.go` all clean. Live-API integration tests run in CI (local env panics on missing `STREAM_KEY` / `STREAM_SECRET`, pre-existing)

Docs side, [docs-content#1276](https://github.com/GetStream/docs-content/pull/1276) now carries the per-SDK error-class table on the Webhooks overview page.

mogita (approved):

JimmyPettersson85 (approved):

